PR #8, opened 2026-04-05
feat(turbovault): fix batch_execute MCP schema with typed input model

slikts:
The loosely-typed batch operations were causing ChatGPT to fail like this when adding the server: "Invalid schema for function 'batch_execute': In context=('properties', 'operations'), array schema items is not an object." I also tested in Claude, but it worked without errors there, so it apparently depends on how permissive the JSON schema parsing is for any particular client.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4431,6 +4431,7 @@ dependencies = [
  "config",
  "criterion",
  "log",
+ "schemars",
  "serde",
  "serde_json",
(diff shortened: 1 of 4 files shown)


nicholasjpaterno:
@slikts thanks for this! I'm merging this as is, I'm going to fast follow with a few small changes and we'll get this shipped asap!

Issue #12, opened 2026-04-08
Bug: query_metadata returns 0 results — Path::ends_with(".md") vs string suffix check


ttfoley:
## Description

`query_metadata` always returns `{"matched": 0}` regardless of vault contents.

## Root cause

`metadata_tools.rs` line 130 uses `file_path.ends_with(".md")` on a `PathBuf`. Rust's `Path::ends_with` does path-component matching, not string suffix matching — so it always returns `false` for `.md` since that's not a valid path component.

```rust
// metadata_tools.rs:130 (broken)
if !file_path.ends_with(".md") {

// export_tools.rs:90 (correct pattern already used elsewhere in the crate)
if !file_path.to_string_lossy().to_lowercase().ends_with(".md") {
```

Quick Rust proof:
```rust
use std::path::PathBuf;
fn main() {
    let p = PathBuf::from("/vault/test.md");
    println!("{}", p.ends_with(".md"));           // false (Path component match)
    println!("{}", p.to_string_lossy().ends_with(".md")); // true (string suffix)
}
```

## Fix

Replace `metadata_tools.rs` line 130:
```rust
// Before
if !file_path.ends_with(".md") {

// After
if !file_path.to_string_lossy().to_lowercase().ends_with(".md") {
```

## Versions affected

1.3.2 (and likely all prior versions)

## Tested

Patched locally, rebuilt, confirmed `query_metadata` returns correct results (0 → 149 matches on a 1033-note vault).

nicholasjpaterno:
Addressed on `main` and shipping in `v1.4.0`. Thanks for this @ttfoley! Feel free to open a PR so we can credit your contributions in the commit history!

PR #14, opened 2026-04-09
test: regression test for #12 (query_metadata .md file matching)

ttfoley:
## Summary

Adds a regression test for #12 — the `Path::ends_with(".md")` bug that caused `query_metadata` to silently return 0 matches.

The existing `test_query_metadata_equality` only asserted the call succeeded (`is_ok()`), not that it actually found matching files. This new test:
- Asserts `matched > 0` for a query that should hit `note1.md`
- Verifies the expected file path appears in results
- Fails with a clear message pointing to #12 if the regression recurs

Didn't want to create more work since you already merged the fix — just tried to contribute something clean and related. Thanks for the invitation and for being so generous with credit!

## Test plan

- [x] `cargo test --package turbovault-tools --test test_metadata_tools test_query_metadata_returns_matches_for_md_files` passes

### DIFF
--- a/crates/turbovault-tools/tests/test_metadata_tools.rs
+++ b/crates/turbovault-tools/tests/test_metadata_tools.rs
@@ -164,6 +164,35 @@ async fn test_query_metadata_equality() {
     let _response = result.unwrap();
 }
 
+/// Regression test for #12: query_metadata must return >0 matches on a vault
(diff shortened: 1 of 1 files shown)


nicholasjpaterno:
Thanks for this @ttfoley!

PR #15, opened 2026-04-18
fix: render inline code correctly in headings, blockquotes, and tables

esumerfd:
## Summary

- **Headings**: Route `Event::Code` to `heading_buffer`/`heading_inline` instead of paragraph buffers, preventing inline code from leaking into the next paragraph
- **Blockquotes**: Re-emit with backtick delimiters into `blockquote_buffer` so content is preserved and re-parseable as inline code
- **Tables**: Re-emit with backtick delimiters into `paragraph_buffer` so cell strings carry inline code markers for downstream renderers
- **Outline (engine.rs)**: Add `Event::Code` arm guarded by `current_heading.is_some()` so inline code text appears in heading outlines

## Test plan
- [ ] Verify inline code renders correctly in headings (not leaked to next paragraph)
- [ ] Verify inline code renders correctly in blockquotes
- [ ] Verify inline code renders correctly in table cells
- [ ] Verify inline code text appears in heading outline extraction

### DIFF
--- a/crates/turbovault-parser/src/blocks.rs
+++ b/crates/turbovault-parser/src/blocks.rs
@@ -510,9 +510,26 @@ fn process_event(event: Event, state: &mut BlockParserState, blocks: &mut Vec<Co
             state.in_strikethrough = false;
         }
         Event::Code(text) => {
-            state.in_code_inline = true;
-            state.add_inline_text(&text);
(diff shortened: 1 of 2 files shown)


nicholasjpaterno:
Thanks for this one @esumerfd

PR #17, opened 2026-04-30
Obsidian Tasks metadata parsing added

ForrestThump:
## Summary

Adds structured parsing for Obsidian Tasks metadata on parsed task items.

This introduces a shared Winnow-based task parser in `turbovault-core` that can read both emoji format and dataview format task metadata.

Unlike the current Obsidian Tasks plugin setting, Turbovault intentionally reads both formats without requiring vault-level configuration.

## What Changed

- Added `TaskItem` fields for structured metadata:
  - created, scheduled, start, due, done, and cancelled dates
  - priority
  - recurrence
  - on-completion behavior
  - task id and dependencies
  - tags, block refs, and custom Dataview metadata
- Added a shared `task_parser` module using `winnow`.
- Integrated metadata parsing into the pulldown-cmark parser path.
- Updated the legacy task line parser to reuse the same metadata parser.
- Added TaskBuilder struct to extract logic from `pulldown_pass()`
- Added compatibility coverage for:
  - Tasks emoji metadata
  - Dataview bracketed and parenthesized fields
  - comma-separated Dataview fields
  - priorities
  - recurrence
  - on-completion metadata
  - task ids and dependencies
  - trailing metadata ordering behavior
  - serde compatibility for older `TaskItem` JSON

## Notes

This does not add new MCP task tools yet. The parsed metadata is now available on `TaskItem`, which should make future task listing/filtering/editing tools easier to build without requiring models or callers to parse raw Markdown themselves.

## Next

- Add / extend task MCP tooling so that the models can easily view, filter, and edit tasks.
    - Including `#hashtag` based filtering like Obsidian Tasks has. I'm thinking it should support more than one hashtag (unlike the plugin). 

## Verification

- `cargo fmt`
- `git diff --check`
- `cargo test -p turbovault-core --lib`
- `cargo test -p turbovault-parser --lib`
- `cargo test`

Please let me know what you think! Happy to adjust future work to feedback. And thank you for Turbovault! It's so useful that I'm extending the task compatibility for selfish reasons.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5000,6 +5000,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tokio",
  "uuid",
+ "winnow",
 ]
 
 [[package]]
@@ -5781,9 +5782,9 @@ checksum = "d6bbff5f0aada427a1e5a6da5f1f98158182f26556f345ac9e04d36d0ebed650"
 
 [[package]]
 name = "winnow"
-version = "1.0.1"
+version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "09dac053f1cd375980747450bfc7250c264eaae0583872e845c0c7cd578872b5"
+checksum = "2ee1708bef14716a11bae175f579062d4554d95be2c6829f518df847b7b3fdd0"
 dependencies = [
  "memchr",
 ]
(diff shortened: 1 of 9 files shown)


ForrestThump:
The existing TaskItem struct has an `is_completed` bool. But maybe that member should be replaced with a `task_status` member of type TaskStatus enum? Because my understanding is that eventually the MCP interface will surface that info to the MCP caller. Thoughts @nicholasjpaterno ?

PR #18, opened 2026-05-02
Hashtag parsing now aligned with Obsidian Tasks plugin behavior.

ForrestThump:
Changed Obsidian task hashtag parsing to align with the behavior of the Obsidian Tasks plugin.

## Before

The parser assumed that `#hashtags` would be in the metadata section after the task description. It also consumed them. That's not actually how the Obsidian Tasks plugin parser works. Hashtags can appear anywhere in the Obsidian Tasks line, and they are parsed but not consumed. The Tasks plugin leaves them in the description field.

## After

The parser now accepts hashtags anywhere after the `- [ ]` portion of the line. It gathers the tags and saves them to the TaskItem `tags` vector and leaves them in the description field. The tags positioning in the metadata is not maintained, but the ordering is. If the tag does not appear in the metadata section it is left in the description untouched. If the tag appears in the metadata section, it is appended to the description in the order it was found. This behavior is aligns with how the Tasks plugin renders the description in Obsidian.

## Gaps

Since the tag position in the metadata section for tags that appear after the description is not recorded, implementing a to_string() method for the TaskItem struct would shift the tags to the lefthand side of the metadata. Probably not an issue since parsing and presenting the task info is the current projected use-case, but if one wanted to edit a file by placing a TaskItem instance as a line, the tags would always be in the description section. However, that seems best from a human-readability standpoint in my opinion.

## Checks

`cargo build`
`cargo test`
`cargo fmt`
`cargo clippy`

All pass.

### DIFF
--- a/crates/turbovault-core/src/task_parser.rs
+++ b/crates/turbovault-core/src/task_parser.rs
@@ -19,6 +19,7 @@
 
 use std::collections::HashMap;
 
+use chrono::NaiveDate;
 use winnow::{
     ModalResult, Parser,
     ascii::space0,
@@ -130,7 +131,8 @@ pub fn parse_task_line(input: &str) -> Result<Task, String> {
 pub fn parse_task_content(content: &str) -> ParsedTaskMetadata {
     let meta_start = find_metadata_start(content);
     let mut parsed = parse_metadata_section(content[meta_start..].trim_start());
-    parsed.description = content[..meta_start].trim_end().to_string();
+    parsed.tags = extract_tags(content);
+    parsed.description = build_description(content, meta_start);
     parsed
 }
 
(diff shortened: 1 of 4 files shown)


ForrestThump:
Switching this to a draft so that I can rebase onto the next release once TurboMCP's tool filtering feature is finished.

nicholasjpaterno:
@ForrestThump just a heads up TurboMCP v3.1.5 is released with the tool filtering support

ForrestThump:
@nicholasjpaterno Do you want to merge the task tooling now (could merge or rebase), or wait for the tool filtering feature? I don't think they really overlap, so it probably wouldn't hurt to implement now. I have been using the task tool calling with Librechat and Openclaw for a couple weeks now and the only issue I had was the nested task parsing, which I think was pre-existing. But I did fix that as well on another branch, just haven't make a PR for it because I wasn't sure if we wanted to tack that onto one of these existing PRs or have it be its own thing.

ForrestThump:
This bug fix is in the task mcp tools PR. No longer needed at all. Closing.

PR #19, opened 2026-05-03
Feature/task mcp tools

ForrestThump:
## Summary

This PR exposes a complete task management interface over MCP, allowing AI agents to list, query, update, and complete Obsidian tasks without touching note structure. It also lays the groundwork for future task tools with a shared write infrastructure that handles hash-gated atomic writes and round-trip markdown serialization.

### New MCP tools

| Tool | What it does |
|------|-------------|
| `list_tasks` | Scan all notes and return tasks with full metadata; filter by status and/or tags |
| `update_task` | Patch any combination of fields on a single task in-place (priority, dates, recurrence, tags, content) |
| `complete_task` | Mark a task done, stamp `done_date`, and spawn the next occurrence for recurring tasks |
| `get_overdue_tasks` | Return all pending tasks whose `due_date` is before today, sorted oldest-first |
| `delete_task` |delete a task based on task id or file + position. This deletes the actual line from the markdown. |

### Supporting infrastructure

**`TaskMetadataFormat` (new type in `turbovault-core`)** — detected at parse time, stored on `TaskItem`, used to round-trip writes without silently converting between emoji (`📅 2026-05-01`) and Dataview (`[due:: 2026-05-01]`) notation.

**`TaskItem::to_markdown_line()` (new method)** — serializes a `TaskItem` back to a canonical Obsidian Tasks markdown line in whichever format was detected. Field order follows Obsidian Tasks convention: `content priority 🔁 🛫 ⏳ 📅 ✅ ❌`.

**`read_task_for_update()` (shared free function in `tools.rs`)** — consolidates the boilerplate shared by `update_task` and `complete_task`: read file → validate hash → split lines (preserving `\r\n` vs `\n`) → bounds-check position → sanity-check checkbox → extract indentation → parse file → locate task. Keeps each tool's body focused on its actual logic.

**`compute_next_occurrence()` (free function)** — recurrence rule parser for the Obsidian Tasks plugin convention. Supports `every [n] day(s|week(s)|month(s)|year(s)|weekday(s)|weekend(s)`. Uses `due_date` as the reference date (falls back to `done_date`) and offsets all three date fields by the same interval so the full scheduling window advances together.

I totally understand if having helpers in `tools.rs` isn't structurally how you want the code organized. Please feel free to suggest moving the helpers to a new `tool_helpers.rs` module if preferred. 

### Correctness properties

- **Hash-gated writes** — every write tool accepts an optional `expected_hash`. If the file changed since the agent last read it, the write is rejected with an actionable error message. No silent overwrites.
- **Atomic writes** — uses the existing `VaultManager::write_file` infrastructure which writes to a temp file and renames.
- **Format-preserving** — emoji tasks stay emoji; Dataview tasks stay Dataview.
- **Indentation-preserving** — leading whitespace is extracted from the original line and re-applied on write, so nested task lists round-trip correctly.

## Test plan

- [x] All 23 integration tests pass
- [x] All 115 `turbovault-core` unit tests pass
- [x] `cargo clippy -- -D warnings` clean
- [x] `cargo fmt --check` clean
- [x] Live-tested end-to-end with OpenClaw agent against a real Obsidian vault. The new tools worked very well :)

**New integration tests added:**
- `test_list_tasks_*` — filtering by status and tag combinations
- `test_update_task_*` — field patching, hash rejection on stale hash, date clearing
- `test_complete_task_spawns_next_occurrence` — recurring task spawns a new pending instance with correct next dates
- `test_complete_task_no_recurrence_preserves_count` — non-recurring completion does not insert extra lines
- `test_get_overdue_tasks_as_of_date` — exact set of overdue tasks for a fixed reference date
- `test_get_overdue_tasks_empty_before_all_due_dates` — returns empty list when no tasks are due yet
- `test_get_overdue_tasks_sorted_oldest_first` — results are in non-decreasing due-date order

All tests use hardcoded `NaiveDate` values and bypass the tool layer so they are fully date-independent.

If the integration tests are too heavy, cluttering, or too much of a maintenance burden, please feel free to let me know and we can consider removing some.

## Gaps

If the calling model doesn't know via system prompt or prompt what tag Obsidian Tasks is filtering for, then if it calls list_tasks without that tag filter, it will return every checkbox item. We could put a "tasks definition" tag to auto filter everything if you want to use an env var or the config yaml file, but I went with the stateless implementation for now. 

Also, the tag filter uses AND logic which makes the most sense to me, but if you did switch it to OR logic, then you'd want the specific Obsidian Tasks single tag "task definition" filter to be ANDed with the OR logic on the other tags.

Anyway, it's stateless now, so we might want to update the docs to tell users to update their system prompts to tell the models to specify that tag filter every time if that's the setting they're using in the Obsidian Tasks plugin. Otherwise it almost looks like a bug even though it's intentional to maintain statelessness.

Actually, idea: since Turbovault obviously has file access, maybe it could automatically retrieve the settings from the Obsidian Tasks settings file in `.obsidian/plugins/obsidian-tasks-plugin/`... that could actually help it automatically align with Obsidian Tasks settings. That might be a good idea for a future PR.

### Convenience Tools

- **`get_tasks_due_today`** — tasks due on the current date
- **`get_tasks_due_this_week`** — tasks due within the current week
- **`get_task_stats`** — aggregate counts by status/priority/tag, overdue count, completion rate; useful for a quick summary without listing every task

### Operation Tools

- **`move_task`** — relocate a task from one note to another (e.g. inbox → project note). Common in GTD-style workflows. Requires coordinated delete-from-source + insert-at-destination.
- **`get_task`** — fetch a single task by `[id:: ...]` dataview field. Precise lookup for tasks that have explicit IDs, avoiding a full `list_tasks` scan.

Please let me know what future tools seem most useful / interesting to add. Obviously the models can already do all this stuff by editing / viewing the lines directly. But the tools will save tokens. I think move_task and delete_task might be most useful for my personal workflow, but I could see someone liking the `get_tasks_due_by` tools. Maybe they should just be one tool that takes a date argument. Also, get_task_stats might be useful to help a model respond well to "prioritize my task list" type queries.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5161,6 +5161,7 @@ name = "turbovault"
 version = "1.5.0"
 dependencies = [
  "anyhow",
+ "chrono",
  "clap",
  "config",
  "criterion",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5086,6 +5086,7 @@ name = "turbovault"
 version = "1.5.0"
 dependencies = [
  "anyhow",
+ "chrono",
  "clap",
  "config",
  "criterion",
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -317,5 +317,45 @@ performance:
 #   security_auditing: true
 
 # ==============================================================================
+# TOOL VISIBILITY
+# ==============================================================================
+# Control which MCP tools are exposed to clients.
+# All name lists accept exact tool names (case-sensitive).
+# Available tags: read, write, delete, admin, graph, search,
+#                 frontmatter, template, batch, export, audit,
+#                 health, semantic, sql
+
+# tool_visibility:
+#   # Whitelist — if set, only these tools are listed and callable.
+#   allowed:
+#     - read_note
+#     - search
+#     - write_note
+#
+#   # Hide from tools/list, still callable by exact name.
+#   hidden:
+#     - full_health_analysis
+#     - vault_quality_report
+#
+#   # Remove from tools/list and reject direct calls.
+#   disabled:
+#     - delete_note
+#     - rollback_note
+#
+#   # Disable ALL tools carrying any of these tags.
+#   disabled_tags:
+#     - delete
+#     - admin
+#
+#   # Hide all tools with these tags from tools/list (still callable by name).
+#   # NOTE: requires turbomcp hide_tags() support (pending); set for forward-compatibility.
+#   hidden_tags:
+#     - semantic
+#     - export
+#
+#   # Hide all tools not annotated read_only = true.
+#   require_read_only: false
+
+# ==============================================================================
 # END OF CONFIGURATION
 # ==============================================================================
--- a/crates/turbovault-core/src/models.rs
+++ b/crates/turbovault-core/src/models.rs
@@ -254,6 +254,10 @@ pub struct TaskItem {
     /// fields remain available here.
     #[serde(default, skip_serializing_if = "HashMap::is_empty")]
     pub metadata: HashMap<String, String>,
+    /// Whether this task's metadata was written in emoji or dataview notation.
+    /// Detected at parse time; used to round-trip writes in the original format.
+    #[serde(default)]
+    pub metadata_format: TaskMetadataFormat,
 }
 
 impl TaskItem {
@@ -264,6 +268,7 @@ impl TaskItem {
         is_completed: bool,
         position: SourcePosition,
     ) -> Self {
+        let metadata_format = TaskMetadataFormat::detect(&parsed.metadata);
         Self {
             content: parsed.description,
             is_completed,
@@ -285,6 +290,96 @@ impl TaskItem {
             tags: parsed.tags,
             block_ref: parsed.block_ref,
             metadata: parsed.metadata,
+            metadata_format,
+        }
+    }
+
+    /// Serialize this task back to an Obsidian Tasks markdown line.
+    ///
+    /// Dispatches to emoji or dataview format based on the format detected at parse time.
+    /// Tags are emitted as part of `content` (where the parser left them).
+    /// Canonical field order: content priority 🔁 🛫 ⏳ 📅 ✅ ❌
+    pub fn to_markdown_line(&self) -> String {
+        match self.metadata_format {
+            TaskMetadataFormat::Emoji => self.to_markdown_line_emoji(),
+            TaskMetadataFormat::Dataview => self.to_markdown_line_dataview(),
+        }
+    }
+
+    fn to_markdown_line_emoji(&self) -> String {
+        let checkbox = if self.is_completed { "[x]" } else { "[ ]" };
+        let content = self.content.trim();
+        let mut meta: Vec<String> = Vec::new();
+
+        let priority_emoji = self.priority.emoji();
+        if !priority_emoji.is_empty() {
+            meta.push(priority_emoji.to_string());
+        }
+        if let Some(ref rec) = self.recurrence {
+            meta.push(format!("🔁 {}", rec));
+        }
+        if let Some(d) = self.start_date {
+            meta.push(format!("🛫 {}", d));
+        }
+        if let Some(d) = self.scheduled_date {
+            meta.push(format!("⏳ {}", d));
+        }
+        if let Some(d) = self.due_date {
+            meta.push(format!("📅 {}", d));
+        }
+        if let Some(d) = self.done_date {
+            meta.push(format!("✅ {}", d));
+        }
+        if let Some(d) = self.cancelled_date {
+            meta.push(format!("❌ {}", d));
+        }
+
+        if meta.is_empty() {
+            format!("- {} {}", checkbox, content)
+        } else {
+            format!("- {} {} {}", checkbox, content, meta.join(" "))
+        }
+    }
+
+    fn to_markdown_line_dataview(&self) -> String {
+        let checkbox = if self.is_completed { "[x]" } else { "[ ]" };
+        let content = self.content.trim();
+        let mut meta: Vec<String> = Vec::new();
+
+        if self.priority != TaskPriority::Normal {
+            let prio_text = match self.priority {
+                TaskPriority::Lowest => "lowest",
+                TaskPriority::Low => "low",
+                TaskPriority::Medium => "medium",
+                TaskPriority::High => "high",
+                TaskPriority::Highest => "highest",
+                TaskPriority::Normal => unreachable!(),
+            };
+            meta.push(format!("[priority:: {}]", prio_text));
+        }
+        if let Some(ref rec) = self.recurrence {
+            meta.push(format!("[recurrence:: {}]", rec));
+        }
+        if let Some(d) = self.start_date {
+            meta.push(format!("[start:: {}]", d));
+        }
+        if let Some(d) = self.scheduled_date {
+            meta.push(format!("[scheduled:: {}]", d));
+        }
+        if let Some(d) = self.due_date {
+            meta.push(format!("[due:: {}]", d));
+        }
+        if let Some(d) = self.done_date {
+            meta.push(format!("[done:: {}]", d));
+        }
+        if let Some(d) = self.cancelled_date {
+            meta.push(format!("[cancelled:: {}]", d));
+        }
+
+        if meta.is_empty() {
+            format!("- {} {}", checkbox, content)
+        } else {
+            format!("- {} {} {}", checkbox, content, meta.join(" "))
         }
     }
 }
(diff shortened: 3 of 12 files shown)


ForrestThump:
Sorry, I think my toolchain might be out of date. I failed linting.

On my machine:

```PowerShell
PS C:\Users\Me\Code\turbovault> cargo clippy --all -- -D warnings
    Checking turbovault-core v1.5.0 (C:\Users\Me\Code\turbovault\crates\turbovault-core)
    Checking turbovault-parser v1.5.0 (C:\Users\Me\Code\turbovault\crates\turbovault-parser)                                                                                                      
    Checking turbovault-audit v1.5.0 (C:\Users\Me\Code\turbovault\crates\turbovault-audit)
    Checking turbovault-graph v1.5.0 (C:\Users\Me\Code\turbovault\crates\turbovault-graph)
    Checking turbovault-export v1.5.0 (C:\Users\Me\Code\turbovault\crates\turbovault-export)
    Checking turbovault-vault v1.5.0 (C:\Users\Me\Code\turbovault\crates\turbovault-vault)                                                                                                        
    Checking turbovault-batch v1.5.0 (C:\Users\Me\Code\turbovault\crates\turbovault-batch)                                                                                                        
    Checking turbovault-sql v1.5.0 (C:\Users\Me\Code\turbovault\crates\turbovault-sql)
    Checking turbovault-tools v1.5.0 (C:\Users\Me\Code\turbovault\crates\turbovault-tools)                                                                                                        
    Checking turbovault v1.5.0 (C:\Users\Me\Code\turbovault\crates\turbovault)                                                                                                                    
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 6.42s                                                                                                                               
PS C:\Users\Me\Code\turbovault> 
```

I will look into this later. Hopefully a simple fix.

---

Now fixed.

ForrestThump:
I discovered during live testing that task parsing doesn't work correctly for indented sub tasks and tasks with indented comments below them. I changed engine.rs to use a stack to track whether it's in an indentation nest and added unit tests to match. 

https://github.com/ForrestThump/turbovault/tree/bug_fix/nested-task-parsing

I still need to conduct live testing on it to verify the fix beyond unit tests. I'm not sure if it would be better to add that fix to this PR, or have it be a separate PR after I verify the fix with a live test. If you'd prefer one PR, I can merge the branch into this branch. Otherwise, I can hold off on submitting the patch until after this PR is merged.

ForrestThump:
Switching this to a draft so that I can rebase onto the next release once TurboMCP's tool filtering feature is finished.

ForrestThump:
Squashed the messy commit history and rebased on the current main branch. Quite sure this is ready now. @nicholasjpaterno

nicholasjpaterno:
Thanks for your patience on this one @ForrestThump.  And, to be clear, the hold has nothing to do with quality on this PR. Everything looks genuinely strong (hash-gated atomic writes, format-preserving round-trips, the recurrence parser, real date-independent tests).
The reason it sat is that it surfaced a decision we needed to make deliberately: what belongs in core vs. what plugs in. This is what I'm thinking and it would make this PR the proving ground.

In principle ideally TurboVault mirrors Obsidian's own structure: a lean core (read/write/graph/search/health), plus an extension model for everything opinionated. The rule that falls out:

▎ If it's a community plugin in Obsidian, it's a plugin in TurboVault.

Obsidian Tasks is itself a community plugin, so task tools belong as a plugin, owned and maintained by you, not folded into core. Thoughts?

I'm also thinking theres a clear distinction between official plugins and community. For a v1 plugin model, I'm leaning toward compiled-in, feature-gated, PR-reviewed Rust providers (plugins = ["tasks"]): clean (no
engine/FFI/sandbox/ABI), safe (everything reviewed before it touches a vault), and it keeps the door open for a potential dynamic JS tier later without rework.

As far as this PR if you're open to the plugin idea, then the mechanics it would depends on don't exist yet, I'm working on landing them first:
- Phase 1: decompose the ~2,700-line tools.rs into per-provider modules composed via turbomcp's CompositeHandler.
- Phase 2: extract a turbovault-plugin-api crate — a curated VaultApi facade + a hook bus (which is exactly what PR #24's vault-event subscriptions become).

Once that scaffold is in, your work lands as the first plugin crate: crates/plugins/turbovault-plugin-tasks, behind a default-off tasks feature, tagged for visibility. The tool logic is unchanged — it moves out of tools.rs into a provider. 

If you agree then I see two easy ways to get there:
1. We land the Phase 1/2 scaffold, then you rebase onto it (lowest friction for you)
2. OR, you start pulling the task tools into a provider module now and we converge.

Either way you'll want to rebase onto current main — the tag-filtering commit this branch was stacked on merged via #25, so a rebase will shrink the diff to just the task work.

Where we land on the open questions

- Namespacing: I'm leaning namespaced (tasks_list_tasks), since it's the MCP-idiomatic disambiguation and makes provenance obvious when several plugins are mounted, and it pairs with the PR #25 tags. You'll own this surface though — if you've got a case for flat names, please let me know.
- Default: off — opt-in like an Obsidian community plugin.
- The statelessness gap you flagged (auto-reading .obsidian/plugins/obsidian-tasks-plugin/ settings): I think that's a strength of the plugin model — a plugin owns its own config and can align with its companion Obsidian plugin. It could be a great follow-up, not a blocker.

Thoughts?!

ForrestThump:
@nicholasjpaterno  All good! I had definitely been thinking about how the coupling could be loosened so that it could have a plugin like system, but I didn't (and don't) really understand the architecture well enough to raise it as an issue.

In regard to maintaining a separate Obsidian Tasks plugin, I'd be happy to do so! Indeed, having a plugin system would also allow me "off the leash" so to speak in the sense that I could just add plugins as I need them for my own workflows, and not really have to worry about integration friction into Turbovault. I wonder if [this](https://github.com/ForrestThump/turbovault/tree/feature/vector-db) could also be a plugin of sorts- even though it's not strictly tied to an Obsidian plugin?

In regard to the plan moving forward, I think I'd prefer that we land the scaffolding first, and then I can work on rebasing the dedicated Tasks crate. I just moved to a new state and started a new job, so I mean I know you're busy too, but I think I'd not have time for good enough quality work for a smooth convergence. Having something like a plugin API to target would help a lot.

Fully agree with the namespacing- I don't really like flatness generally. Gets too confusing.

Absolutely concur with default OFF note. Note that my vector DB feature branch on my fork also disables that feature since it adds performance overhead. I like the compile time plugin linking approach via Rust features instead of doing some sort of settings / config file schema. I know most programs that use plugins work that way, but it's not very Rusty or high performance. 

Yep, also agree with the advantages you cited about the plugin model, and I think it kind of boils down to loosened coupling. We wouldn't have to worry about Turbovault accessing `.obsidian/plugins/`, the plugin crates could handle it themselves, although we might want to extract common logic / duplication into a plugin library crate at some point. Probably not now, but after we get 2-3 plugins up and running, the overlap / duplicate code would probably be pretty apparent for extraction.

Also, full disclosure, I'm not new to Rust or big projects, but I am new to FOSS development and Rust crate publishing. Once I get the plugin set up, should it be a crate published on crates.io, or would the optional feature dependencies in the Turbovault toml(s) just point to a dedicated Github repo (that I would create)? I understand that both are probably possible, but what would be preferred / idiomatic?

nicholasjpaterno:
Keeping this PR open as the reference implementation for the Tasks vertical. The architecture underneath it has changed since this diff was authored: Phase 1 is now complete on main in abea1ec/#28, with the former tools.rs split into focused providers behind a flat compatibility facade.\n\nFor this PR to become merge-compatible, the intended path is:\n\n1. Wait for or help define Phase 2 in #34: the curated turbovault-plugin-api/VaultApi facade plus hook contract.\n2. Rebase onto current main and drop the already-landed tag-filtering/tool-visibility history.\n3. Move the Tasks implementation out of core tools.rs into crates/plugins/turbovault-plugin-tasks.\n4. Gate it behind a default-off tasks Cargo feature and expose namespaced tools such as tasks_list_tasks, tasks_update_task, and tasks_complete_task.\n5. Keep task-specific configuration and Obsidian Tasks integration inside that plugin; only genuinely format-level parsing/model primitives should remain in shared core crates.\n6. Adapt the existing tests to run with the tasks feature, including visibility, namespace, hash-gated writes, recurrence, and format-preserving round trips.\n\nNo need to churn through a rebase before the Phase 2 facade is concrete unless you want to help shape that contract. The current PR remains useful evidence and a solid behavioral specification for the eventual plugin.

nicholasjpaterno:
Phase 2 now has a concrete implementation in #39. This PR remains open as the task behavior reference.

Compatibility path after #39 lands:
1. extract the vertical under crates/plugins/turbovault-tasks;
2. add a default-off plugin-tasks feature including plugin-api;
3. implement Plugin and PluginProvider with local tool names;
4. use VaultApi for note listing, reads, create-only writes, and version-matched writes instead of raw server/manager access;
5. advertise the resulting tools only as tasks_*;
6. port the behavioral tests onto the new provider contract.

A task crate may also expose a broader direct Rust API; the mounting contract does not prohibit that. No merge/rebase is requested yet because the current diff still targets the pre-provider architecture.

ForrestThump:
Superseding this in favor of a rebuild on the stable plugin API.

When I opened this, the task tools were written against the old manager-based `tools.rs`. Since the plugin API landed (#39), I've rebuilt the entire task surface as a compiled-in, default-off **module** (`turbovault-plugin-tasks`) against the curated `VaultApi` — CAS-only writes, namespaced `tasks_*` tools — instead of extending the monolith. It covers everything here (`list_tasks` / `get_overdue_tasks` / `list_task_tags` / `complete_task` / `update_task` / `delete_task`) and adds self-tuning to the user's Obsidian Tasks settings (emoji vs dataview format, global filter) and recurrence-spawn on complete.

New branch: https://github.com/ForrestThump/turbovault/tree/feat/plugin-tasks

It builds on the small read-only `read_config` capability in #41; once that lands I'll rebase and open the module as its own PR. Closing this to avoid confusion in the meantime.

PR #23, opened 2026-05-18
perf: cache-first query_metadata with mtime validation (10.6× faster)

ForrestThump:
## Problem

`query_metadata` called `scan_vault()` followed by `parse_file()` on every single
invocation. For a 100-note vault this issued ~100 `statx` syscalls (directory scan)
plus ~100 file reads and YAML parses on every call, regardless of whether any file
had changed since the last query.

`VaultManager` already maintains a fully-populated `Arc<RwLock<HashMap<PathBuf,
CacheEntry>>>` that is initialized during `initialize()` and kept up-to-date on every
write, delete, and move operation. `query_metadata` never used it.

### Before (abbreviated)

```rust
pub async fn query_metadata(&self, pattern: &str) -> Result<Value> {
    let files = self.manager.scan_vault().await?;    // full dir scan
    for file_path in files {
        match self.manager.parse_file(&file_path).await {  // read + YAML parse per file
            Ok(vault_file) => { /* filter */ }
```

## Fix

Added `VaultManager::vault_files_validated()` which reads from the in-memory cache and
re-reads only files whose `mtime` is newer than their `cached_at` timestamp, catching
external modifications (Obsidian, editors, git sync) without a full vault re-scan.

Updated `query_metadata` to call `vault_files_validated()` instead of
`scan_vault()` + `parse_file()`.

### Key implementation detail: batched stat calls

The mtime check batches all `stat` syscalls into a single `tokio::task::spawn_blocking`
call rather than issuing N individual `tokio::fs::metadata` futures. Each async
`tokio::fs::metadata` carries task-scheduling overhead; for a 100-file vault that
overhead accumulated to ~900 µs. Batching all N stat calls into one `spawn_blocking`
task (using synchronous `std::fs::metadata`) reduces that to a single task dispatch.

### Two-phase locking

The implementation avoids holding any lock during I/O:

1. **Read lock** — snapshot `(path, cached_at)` pairs into a `Vec`; drop lock.
2. **No lock** — all `stat` calls in `spawn_blocking`; re-read and re-parse stale files.
3. **Write lock** — update stale entries; remove deleted files; drop lock.
4. **Read lock** — collect and return the validated cache.

## Benchmark results

Measured on a 100-note vault with frontmatter. Criterion, 100 samples each.
Machine: x86-64 Linux, local ext4 filesystem.

| Variant | Median | vs. before |
|---|---|---|
| before — `scan_vault()` + `parse_file()` per call | 3.44 ms | baseline |
| after — `vault_files_validated()` (mtime-validated cache) | 324 µs | **10.6× faster** |

Hot path (no files changed between calls): one `spawn_blocking` task issuing N
synchronous `stat` calls, then one read lock to return cached `VaultFile` clones.
Zero file reads on the hot path.

Worst case (every file changed): equivalent to the old scan-on-every-call cost,
plus the one `spawn_blocking` dispatch overhead (~50 µs).

### A/B bench

The `metadata_query_ab` criterion group in `tool_benchmarks.rs` reproduces these
measurements side-by-side. Requires `[[bench]] harness = false` in
`crates/turbovault/Cargo.toml` (see [fix/criterion-harness](https://github.com/ForrestThump/turbovault/tree/fix/criterion-harness)).

## Correctness

- Files modified externally are detected via `mtime` comparison and re-parsed before
  the query runs. The window between "mtime check" and "re-parse" is closed by the
  fact that we re-read the file content immediately after detecting staleness.
- Files deleted externally are detected when `std::fs::metadata` returns an error
  (`ENOENT`); their cache entries are removed in Phase 3.
- Files written via `VaultManager::write_file()` already update `cached_at` on the
  cache entry, so they will never be flagged as stale by this check.
- Concurrent writes during a query could cause a re-parsed entry to be written back
  slightly stale. This is the same eventual-consistency window that exists in the
  original implementation and is acceptable for a read-query path.

## Files changed

- `crates/turbovault-vault/src/manager.rs` — added `vault_files_validated()`,
  removed unused `is_file_modified_since()` (logic inlined into `spawn_blocking` closure)
- `crates/turbovault-tools/src/metadata_tools.rs` — `query_metadata` now delegates to
  `vault_files_validated()` instead of `scan_vault()` + `parse_file()`
- `crates/turbovault/benches/tool_benchmarks.rs` — updated `metadata_query_ab` group
  to compare before/after inline

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -317,5 +317,45 @@ performance:
 #   security_auditing: true
 
 # ==============================================================================
+# TOOL VISIBILITY
+# ==============================================================================
+# Control which MCP tools are exposed to clients.
+# All name lists accept exact tool names (case-sensitive).
+# Available tags: read, write, delete, admin, graph, search,
+#                 frontmatter, template, batch, export, audit,
+#                 health, semantic, sql
+
+# tool_visibility:
+#   # Whitelist — if set, only these tools are listed and callable.
+#   allowed:
+#     - read_note
+#     - search
+#     - write_note
+#
+#   # Hide from tools/list, still callable by exact name.
+#   hidden:
+#     - full_health_analysis
+#     - vault_quality_report
+#
+#   # Remove from tools/list and reject direct calls.
+#   disabled:
+#     - delete_note
+#     - rollback_note
+#
+#   # Disable ALL tools carrying any of these tags.
+#   disabled_tags:
+#     - delete
+#     - admin
+#
+#   # Hide all tools with these tags from tools/list (still callable by name).
+#   # NOTE: requires turbomcp hide_tags() support (pending); set for forward-compatibility.
+#   hidden_tags:
+#     - semantic
+#     - export
+#
+#   # Hide all tools not annotated read_only = true.
+#   require_read_only: false
+
+# ==============================================================================
 # END OF CONFIGURATION
 # ==============================================================================
--- a/crates/turbovault-core/src/cache.rs
+++ b/crates/turbovault-core/src/cache.rs
@@ -78,6 +78,30 @@ impl VaultCache {
         })
     }
 
+    /// Initialize cache with an explicit cache directory, bypassing `get_cache_dir()`.
+    ///
+    /// Use this in tests instead of mutating `XDG_CACHE_HOME`. Passing the cache
+    /// dir directly is inherently thread-safe; env-var mutation is not.
+    pub async fn init_with_cache_dir(cache_dir: &Path, project_root: &Path) -> Result<Self> {
+        let project_id = Self::get_project_id(project_root)?;
+        let project_cache_dir = cache_dir.join("projects").join(&project_id);
+
+        if !project_cache_dir.exists() {
+            fs::create_dir_all(&project_cache_dir)
+                .await
+                .map_err(Error::io)?;
+        }
+
+        Ok(Self {
+            cache_dir: cache_dir.to_path_buf(),
+            project_cache_dir: project_cache_dir.clone(),
+            vaults_file: project_cache_dir.join("vaults.yaml"),
+            metadata_file: project_cache_dir.join("metadata.json"),
+            project_id,
+            working_dir: project_root.to_path_buf(),
+        })
+    }
+
     /// Initialize cache with a specific project (useful for testing)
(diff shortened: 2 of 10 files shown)


nicholasjpaterno:
Thanks for this @ForrestThump! I've finally had a chance to go through this and the write-up/benchmarking are much appreciated!

I found one issue before merging: `vault_files_validated()` only validates paths already present in `file_cache`, but `write_file()` currently removes the cache entry and does not reinsert the newly parsed `VaultFile`. `move_file()` similarly removes the old path but does not cache the new one. That means after a normal MCP write/edit/move, `query_metadata` can silently miss that note until a full reinitialize.

I think this is very close. Could you update the cache after successful `write_file()` / `move_file()`? If you have a chance to add a regression test for querying metadata after a write or move without calling `initialize()` again - I can also add this after merge. We can handle broader external-new-file detection separately if needed, but I think the internal mutation path should be covered.

Really appreciate all the contributions @ForrestThump! This is a good perf win once that cache consistency edge is closed.

ForrestThump:
@nicholasjpaterno Thanks for catching that! I'll add test coverage for those cases and then fix the regression. 

Side note: Do you think the symlink gap / regression is worth fixing? Basically it would require adding a `is_symlink` boolean to CacheEntry, and then the staleness check would need to always consider symlinks stale. `d_type` doesn't return the path of the symlink, so you'd basically have to call `readlink` + `stat` for each entry that is flagged as a symlink. 

You could save the exact path instead of flagging it, but that would break the point of the symlink and potentially confuse the few users that use symlinks for a marginal speed gain on symlinks... so the `readlink` + `stat` approach for flagged entries would be best. It's just so niche that I didn't bother + the added code complexity might not be worth it.

For all I know, symlinks might not even work with Obsidian (honestly not sure). So that use case might already be filtered out. Might have to test / research that more. Or we just wait until someone opens it as an issue. :P

ForrestThump:
@nicholasjpaterno Fixed regression, added more test coverage, reduced code duplication via test helpers. Also cleaned up commit history. I also rebased off of the tool filtering by tag feature commit to make sure it could merge in cleanly if that PR is merged to `main` first... but let me know if you prefer PRs to be strictly branched off of `main` in the future.

Sorry, there is kind of a meta-comment on line 577 of test_metadata_tools.rs. But it might be nice because it makes it clear which tests would flag the regression.

PR #24, opened 2026-05-20
yolo

ahuserious:


ForrestThump:
Analyzed and isolated.

The subscription feature is cool. Not something that I'd personally use I don't think... not sure that it's worth the rebase work. At least it's not something that I would do.

Claude Code identified a pre-existing bug + fix from the PR though:

---

The Bug
  
  VaultEvent::FileRenamed is defined but can never be emitted.

  In upstream watcher.rs:176, convert_event matches on EventKind::Modify(_) — a wildcard that catches everything including rename events. When a file is renamed, notify fires
  EventKind::Modify(ModifyKind::Name(...)), which falls into that arm and comes out as VaultEvent::FileModified. There is no arm that constructs FileRenamed.

  The variant has existed in the enum since at least the initial commit, complete with its docstring and a path() impl — it just silently never fires.

  Why It's Nontrivial to Fix

  The fix isn't just adding a match arm. notify delivers renames differently per platform:

  - Linux: one event, ModifyKind::Name(RenameMode::Both), with both paths in a single event.paths vec
  - macOS / Windows: two separate events — RenameMode::From (old path), then RenameMode::To (new path) — with one path each

  The current convert_event takes a single Event with no state between calls, so it has no way to pair the macOS/Windows split events. The fix requires a correlation buffer that holds the From half and waits for the To, with a timeout fallback to emit FileDeleted if the To never arrives.

  Practical Impact

  Any code consuming VaultEvent that has a FileRenamed arm is dead code on the upstream. A rename always surfaces as FileModified(new_path) — the old path disappears with no event, and the new path looks like an edit to a pre-existing file. Downstream things like the subscription registry, any index that tracks file identity, or anything that cares about continuity of a note's history will silently get it wrong.

---

So that's good to know. Maybe the CorrelationBuffer architecture in `watcher.rs` would be worth using?

nicholasjpaterno:
Keeping this PR open as the reference implementation for vault-event subscriptions and pull delivery. Its current diff targets the former monolithic tools.rs and an older vault/provider lifecycle, so it is not merge-ready against current main.

The compatibility path after #28 is:
1. Separate the core watcher correctness fix (cross-platform rename correlation and stranded-rename handling) from the subscription vertical so it can be reviewed independently.

2. Target the Phase 2 contract in #34: shared event/hook types belong in turbovault-plugin-api, while raw VaultManager and server internals stay behind VaultApi.
3. Move subscription registry and MCP tools into a default-off vault-events plugin/provider, with namespaced tools.
4. Rework lifecycle around the active-vault/provider context rather than pinning the first selected vault for the server lifetime.
5. Preserve the pull-based sequence/gap semantics and bounded queues, then rerun the stdio end-to-end test through the new provider facade.
6. Rebase onto current TurboMCP and current main only after that facade is concrete; the existing diff remains useful as the behavioral and testing reference.

The provenance questions are tracked in #33. This comment is an architecture update, not a rejection of the feature.

nicholasjpaterno:
Phase 2 now has a concrete implementation in #39. This PR remains open as the subscription behavior reference.

Compatibility path after #39 lands:
1. extract the watcher/subscription vertical under crates/plugins/;
2. add a default-off vault-events feature including plugin-api;
3. translate watcher output into the shared HookBus envelope instead of defining a second event contract;
4. preserve bounded delivery and make lag require an authoritative VaultApi resync;
5. expose subscription tools only under a vault_events_* namespace;
6. treat provenance as best-effort and ExternalOrUnknown as fail-open, never as auth.

The per-subscriber long-poll/filter/cursor behavior in this PR remains useful reference material, but the implementation needs to be rebuilt on the current provider and hook boundaries rather than rebased as a core diff.

PR #31, opened 2026-06-21
fix: Concurrency hardening. Hash check prior to file edit added to VaultManager::edit_file

ForrestThump:
# Concurrency improvements: edit_file TOCTOU fix + per-operation optimistic concurrency in batches

## Summary

Two concrete hardening improvements to Turbovault's existing optimistic-concurrency support
(`expected_hash` + atomic rename), plus a clearly-scoped **maintainer decision** on the conflict
error type that is deliberately left out of the shipped diff to avoid a breaking change.

1. **`edit_file` TOCTOU fix** — `edit_file` now re-checks the file's content hash at write time
   instead of writing unconditionally after dropping its cache lock. A concurrent write that lands
   between the read and the atomic rename is now detected and rejected instead of silently
   overwriting the other writer's change.
2. **Per-operation `expected_hash` in `batch_execute`** — `WriteNote`, `DeleteNote`, and `MoveNote`
   batch operations accept an optional `expected_hash`, verified in a pre-flight pass before any
   operation is applied. Wire-compatible (JSON unchanged for existing payloads).

This PR is intentionally **non-breaking at the public-enum level**: it does **not** add or remove any
`Error` variant. A richer structured conflict error is described below as an opt-in the maintainer
can choose to adopt (see [API decision](#api-decision-conflict-error-shape)).

## Motivation

Turbovault already exposes optimistic concurrency end-to-end (`read_note` returns a `hash`;
`write_note`/`edit_note`/`delete_note`/`move_note` accept `expected_hash`, checked before the write).
Two gaps remained for concurrent human + multi-agent use:

- `edit_file` validated the hash under a lock, then **released the lock and wrote with no re-check**,
  leaving a window where a concurrent write could be silently lost.
- Batches had no way to express "apply only if these files are still at the hashes I last read,"
  which coordinated multi-file agent edits need.

### 1. `edit_file` TOCTOU fix — [`crates/turbovault-vault/src/manager.rs`]

Previously `edit_file` did: lock → read → validate `expected_hash` → **drop lock** →
`write_file(.., None)`. Because the final write passed `None`, it performed no concurrency check, so
a write landing between the read and the rename was silently clobbered.

Now `edit_file` computes the hash of the content it actually read (`validated_hash`) and forwards it
to the write:

```rust
let validated_hash = compute_hash(&current_content);
// ... validate caller's expected_hash (if any) against validated_hash ...
// ... apply edits ...
self.write_file(&vault_path, &new_content, Some(&validated_hash)).await?;
```

If the file changed since the engine read it, the write now fails with `ConcurrencyError` rather
than overwriting. This is purely internal — **no signature change** — and behaves as a bug fix: the
only newly-observable behavior is a conflict error in a race that previously produced a lost update.

> Scope note: atomic-rename means this narrows the window to the same minimal read→rename window all
> single writes share; it does not (and cannot) mutually exclude **external** editors such as
> Obsidian writing directly. Those are *detected* (hash mismatch → error), not *prevented*, which is
> the correct guarantee for an optimistic scheme.

### 2. Per-operation `expected_hash` in batches — [`crates/turbovault-batch/src/lib.rs`]

`WriteNote`, `DeleteNote`, and `MoveNote` gain an optional `expected_hash`:

```rust
WriteNote { path: String, content: String, #[serde(default)] expected_hash: Option<String> }
DeleteNote { path: String, #[serde(default)] expected_hash: Option<String> }
MoveNote { from: String, to: String, #[serde(default)] expected_hash: Option<String> }
```

A new pre-flight pass (`precheck_preconditions`) verifies **all** preconditions before **any**
operation runs, so a stale hash fails the whole batch without partial application:

```rust
self.validate(&ops).await.and(self.precheck_preconditions(&ops).await)
```

`CreateNote` and `UpdateLinks` carry no precondition. The `batch_execute` MCP tool description and
examples were updated to mention the field; the JSON schema gains an optional field via
`#[serde(default)]`, so **existing batch payloads deserialize unchanged**.

## Tests

All new behavior is covered; full workspace suite green, clippy clean.

- `edit_file` stale hash → conflict; success path; **TOCTOU re-check rejects an intervening write**
- `write_file` stale hash → conflict; `expected_hash` on a missing file → conflict explaining the
  file is gone
- `delete_file` / `move_file` stale hash → conflict, target left untouched
- batch: matching `expected_hash` succeeds; **stale precondition aborts in pre-flight**
  (`executed == 0`, no side effects); `DeleteNote` + `MoveNote` precondition arms covered

(The TOCTOU test simulates the interleave deterministically at the `write_file` layer — driving the
write with the now-stale pre-image hash that `edit_file` forwards internally — since a true in-call
race isn't deterministically injectable without a test hook.)

## API / versioning impact

| Change | Wire (MCP/JSON) | Rust source | Notes |
|---|---|---|---|
| `edit_file` TOCTOU fix | compatible | compatible | internal only; behavioral bug fix |
| batch `expected_hash` fields | compatible (`#[serde(default)]`) | **source-break for direct construction** | see below |
| conflict error shape | compatible | compatible (unchanged) | structured variant offered as opt-in below |

**Batch field caveat:** adding fields to the public `BatchOperation` struct-variants is source-
breaking for any **Rust** code that constructs those variants as literals (Rust struct literals
require all fields). It is **not** wire-breaking. In practice `BatchOperation` is built by
deserializing MCP tool JSON, not constructed by hand in downstream Rust, so real-world impact is
low. There is no way to add a per-operation field in Rust without this; if the maintainer wants
*zero* Rust source-break, the alternative is separate variants
(`WriteNoteChecked { .., expected_hash }`), which is uglier and not recommended.

## <a name="api-decision-conflict-error-shape"></a>API decision: conflict error shape

This is the one place where there's a real "pick one" choice, so it's surfaced rather than decided.

### Option A — keep the string-based `ConcurrencyError` (what this PR ships)

Conflicts continue to return the existing `Error::ConcurrencyError { reason: String }`, with the
hashes interpolated into the message. **No change to the public `Error` enum**, so this option needs
no semver consideration on that axis.

- ✅ Non-breaking to the `Error` enum.
- ➖ Callers (especially agents) that want to recover programmatically must parse the message string
  to recover the expected/actual hashes.

### Option B — add a structured `ConcurrentModification` variant (opt-in)

Replace the string for these four call sites with a typed variant carrying the data needed for
machine recovery:

```rust
/// Concurrent modification detected by an optimistic-concurrency hash check.
/// `actual` is `None` when the file no longer exists.
#[error("Concurrent modification of {path}: expected hash {expected}, found {actual:?}. \
         Re-read the file and retry.")]
ConcurrentModification { path: PathBuf, expected: String, actual: Option<String> },
```

`write_file` / `edit_file` / `delete_file` / `move_file` return this instead of `ConcurrencyError`;
the MCP layer surfaces `expected` / `actual` / `path` as fields so agents don't parse strings.

- ✅ Machine-readable conflict signal — the most useful form for agent-driven retry/merge/escalate.
- ➖ **Public-enum change.** `Error` is not `#[non_exhaustive]`, so adding a variant breaks downstream
  exhaustive `match`es, and switching the returned variant is a behavioral break for anyone matching
  `ConcurrencyError`. Under strict semver this is a breaking change (→ `2.0.0`), though additive
  enum changes are often treated as minor in practice — the maintainer's call.

> If Option B is preferred, I'm happy to push it as a follow-up commit on this branch (the variant +
> the four call sites + structured field assertions in the existing conflict tests are
> straightforward). The implementation already exists in history and can be restored quickly.

## Out of scope

- A higher-level `edit_note_with_retry` helper (easy follow-up; the read→edit→conditional-write loop
  is already expressible).
- Distributed/advisory locking, automatic merge, and provenance — tracked separately.

### DIFF
--- a/crates/turbovault-batch/src/lib.rs
+++ b/crates/turbovault-batch/src/lib.rs
@@ -92,10 +92,12 @@
 //! let write = BatchOperation::WriteNote {
 //!     path: "file.md".to_string(),
 //!     content: "content".to_string(),
+//!     expected_hash: None,
 //! };
 //!
 //! let delete = BatchOperation::DeleteNote {
 //!     path: "file.md".to_string(),
+//!     expected_hash: None,
 //! };
 //!
 //! assert!(write.conflicts_with(&delete));
@@ -145,15 +147,35 @@ pub enum BatchOperation {
 
     /// Write/overwrite a note
     #[serde(rename = "WriteNote", alias = "WriteFile")]
-    WriteNote { path: String, content: String },
+    WriteNote {
+        path: String,
+        content: String,
+        /// Optional optimistic-concurrency precondition: SHA-256 hash the file
+        /// is expected to have before the write. Checked before the operation
+        /// is applied; a mismatch fails the batch with a ConcurrencyError.
+        #[serde(default)]
+        expected_hash: Option<String>,
+    },
 
     /// Delete a note
     #[serde(rename = "DeleteNote", alias = "DeleteFile")]
-    DeleteNote { path: String },
+    DeleteNote {
+        path: String,
+        /// Optional optimistic-concurrency precondition (see `WriteNote`).
+        #[serde(default)]
+        expected_hash: Option<String>,
+    },
 
     /// Move/rename a note
     #[serde(rename = "MoveNote", alias = "MoveFile")]
-    MoveNote { from: String, to: String },
+    MoveNote {
+        from: String,
+        to: String,
+        /// Optional optimistic-concurrency precondition on the source file
+        /// (see `WriteNote`).
+        #[serde(default)]
+        expected_hash: Option<String>,
+    },
 
     /// Update links in a note (find and replace link target)
     #[serde(rename = "UpdateLinks")]
@@ -170,8 +192,8 @@ impl BatchOperation {
         match self {
             Self::CreateNote { path, .. } => vec![path.clone()],
             Self::WriteNote { path, .. } => vec![path.clone()],
-            Self::DeleteNote { path } => vec![path.clone()],
-            Self::MoveNote { from, to } => vec![from.clone(), to.clone()],
+            Self::DeleteNote { path, .. } => vec![path.clone()],
+            Self::MoveNote { from, to, .. } => vec![from.clone(), to.clone()],
             Self::UpdateLinks {
                 file,
                 old_target,
@@ -179,6 +201,31 @@ impl BatchOperation {
             } => {
                 vec![file.clone(), old_target.clone(), new_target.clone()]
             }
+        }
+    }
+
+    /// Optimistic-concurrency precondition for this operation, if any.
+    ///
+    /// Returns `(path, expected_hash)` where `path` is the file whose current
+    /// content hash must equal `expected_hash` for the operation to proceed.
+    /// `CreateNote` and `UpdateLinks` never carry a precondition.
+    pub fn precondition(&self) -> Option<(&str, &str)> {
+        match self {
+            Self::WriteNote {
+                path,
+                expected_hash: Some(h),
+                ..
+            } => Some((path, h)),
+            Self::DeleteNote {
+                path,
+                expected_hash: Some(h),
+            } => Some((path, h)),
+            Self::MoveNote {
+                from,
+                expected_hash: Some(h),
+                ..
+            } => Some((from, h)),
+            _ => None,
         }
     }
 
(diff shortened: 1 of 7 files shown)


ForrestThump:
This one is pretty straightforward. I would guess that end users haven't used Turbovault much with concurrent agents, and the gap between hash check and write time is narrow enough that data contention wasn't a big isue. I am leaving it as a draft for now to see if you want to have a `ConcurrentModification` error type, or just keep the more generic `ConcurrencyError` with string description.

nicholasjpaterno:
Integrated on main in 877c295, with co-author attribution to Shiloh. The landed version preserves both core improvements from this PR: edit_file now forwards its validated pre-image hash into the final atomic write, and batch WriteNote/DeleteNote/MoveNote operations accept expected_hash values that are all preflighted before any operation runs. I adapted the MCP schema and provider description to the decomposed architecture and added manager, batch, catalog, and public-facade regression coverage. Full result: 879 unit/integration tests plus doctests pass; strict all-target/all-feature Clippy is clean. Thank you for the careful API analysis and implementation.

Issue #34, opened 2026-07-05
[Tracking] Plugin & extensibility architecture


nicholasjpaterno:
Umbrella issue tracking the plugin and extensibility architecture. The goal: keep the core a lean Obsidian-vault SDK (read/write/graph/search/health) while opinionated feature verticals plug in as feature-gated, PR-reviewed providers — mirroring Obsidian's own core-vs-community-plugin split.

Guiding principle: **if it's a community plugin in Obsidian, it's a plugin in TurboVault.** Task management, vault-event subscriptions, and vector search are verticals, not core.

### v1 model
Compiled-in, Cargo-feature-gated Rust providers under `crates/plugins/`, each depending on a shared `turbovault-plugin-api` contract crate (a `VaultApi` facade + provider trait + hook/event types). No dynamic engine, no FFI, no sandbox — every plugin is reviewed before it ships. A future dynamic (JS) tier stays a back-pocket option, gated on real demand, and would target the same facade.

### Phased plan

- [x] **Phase 1 — Provider decomposition** (#28). Split `tools.rs` into composable providers via turbomcp's `CompositeHandler`. Pure refactor, no public tool renames, tests prove parity. Foundation for everything else.
- [x] **Phase 2 — Extract `turbovault-plugin-api`.** The `VaultApi` facade over internals (never raw access) plus the hook/event bus that the #24 vault-event work becomes. Valuable as plain layering even with zero plugins.
- [ ] **Phase 3 — First feature-gated verticals**, each under `crates/plugins/`, default-off, namespaced tools:
  - [ ] Tasks (#19)
  - [ ] Vault-event subscriptions (#24)
  - [ ] Vector search (#29)
- [ ] **Phase 4 (deferred) — Dynamic JS tier.** Only on demonstrated external demand: a JS engine over the same facade plus Obsidian-style enable/disable UX.

### Conventions
- Core tools stay flat (`read_note`); plugin tools are namespaced (`tasks_list_tasks`).
- Feature-gated verticals default **off**.
- Plugin authors maintain their crate; we keep the merge button (curated, not abandoned).
- Shared plugin-helper extraction is deferred until 2-3 plugins make the overlap concrete.

Open to input from anyone — @ForrestThump, your task and vector-db branches are the proving ground for Phases 2-3, and you're very welcome to drive the pieces you have bandwidth for.

ForrestThump:
[...]

Issue #28, opened 2026-06-06
[Phase 1] Decompose tools.rs into composable providers


nicholasjpaterno:
The ~2,700-line `tools.rs` holds all ~64 tools in one `#[server]` struct. Split it into focused provider handlers, composed via turbomcp's `CompositeHandler`, as the foundation for the plugin model (see tracking issue).

**Hard constraint:** public tool names must not change. `CompositeHandler` namespaces mounted handlers as `{prefix}_{tool}`, so **core must stay flat** (empty-prefix / flat-merge) — `read_note` stays `read_note` so existing Claude Desktop / client configs keep working. Namespacing (`tasks_*`) is reserved for optional verticals.

### Tasks
- [ ] Verify flat/empty-prefix mounting for core (or add it) so no public tool is renamed
- [ ] Group tools into provider modules (graph / search / metadata / templates / audit / vault-lifecycle / …)
- [ ] Parity test: `tools/list` output identical before/after

Pure refactor, no behavior change. Core-team task; sequenced before any vertical lands.

ForrestThump:
I ran into [this blocker](https://github.com/Epistates/turbomcp/issues/22) for the flatmapping. Grok suggested a handler in Turbovault to handle tool naming schema, but it seems like a change to TurboMCP makes more sense. I wouldn't want to reinvent the wheel on other MCPs with the desired naming schema.

Also do we want to keep #36 as draft and rename it and add on the #28 refactor once the flat naming system is solidified (like whether it lives in TurboMCP or a Turbovault handler), or merge it in now? I lean toward leaving it as a draft, changing the name to `Decompose tools.rs...` , and then committing the decomposition on the same PR.

nicholasjpaterno:
Completed on main in abea1ec. The former monolithic tool surface is split into 13 focused providers composed through CompositeHandler, while the public core API remains flat through the compatibility facade. The checked-in pre-split catalog and regression tests verify byte-for-byte tools/list parity, dispatchability of every advertised tool, public names, schemas, tags, and shared state. The full all-feature workspace suite passes (871 unit/integration tests plus doctests), and strict Clippy is clean. Closing as complete.

PR #36, opened 2026-07-15
test: pin tools/list parity for provider decomposition (#28)

ForrestThump:
## Summary

Adds a regression suite that pins the public MCP tool surface before the Phase 1
`tools.rs` → composable providers refactor (#28), foundation for the plugin /
extensibility plan (#34).

Existing tests only sampled a few tool names or exercised lower-level tool crates.
They would not catch renames, dropped tools, tag/annotation loss, schema property
drift, or CompositeHandler prefix leaks (`graph_read_note` instead of `read_note`).

## What this locks

- Exact `tools/list` name set (64 core tools)
- Tags + `readOnlyHint` / `destructiveHint` (visibility filtering)
- Input schema property names (client parameter contracts)
- Flat core naming (no accidental provider prefixes; namespacing reserved for plugins)
- `call_tool` dispatch for every listed tool (ToolNotFound is a hard failure)

When intentionally adding/removing/renaming a core tool, update `EXPECTED_TOOLS`
in `crates/turbovault/tests/test_tools_list_parity.rs` in the same PR.

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --workspace --all-features --all-targets -- -D warnings`
- [x] `cargo test --workspace --all-features`
- [x] `cargo test -p turbovault --test test_tools_list_parity`

Closes nothing by itself — enables safe implementation of #28.
Related: #28, #34

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -923,9 +923,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-epoch"
-version = "0.9.18"
+version = "0.9.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5b82ac4a3c2ca9c3460964f020e1402edd5753411d7737aa39c3714ad1b5420e"
+checksum = "2d6914041f254d6e9176c01941b21115dcfb7089e55135a35411081bd106ef3f"
 dependencies = [
  "crossbeam-utils",
 ]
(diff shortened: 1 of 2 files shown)


nicholasjpaterno:
Thanks @ForrestThump — the parity concern this PR identified is now covered on main as part of abea1ec, alongside the completed provider decomposition. The landed suite checks byte-for-byte equivalence against the checked-in pre-split catalog, flat public names, schemas/tags/annotations, dispatchability of every advertised tool, resource routing, and shared provider state. Since the implementation and broader regression coverage are now on main, this draft is superseded. Your framing of the compatibility constraints directly informed the landed tests.

Issue #33, opened 2026-07-02
Write Provenance in the Vault Event Stream — Design Discussion


ForrestThump:
**Author**: ForrestThump
**Date**: July 1, 2026
**Status**: Discussion / RFC — seeking maintainer direction
**Related**:
- PR [#24](https://github.com/Epistates/turbovault/pull/24) (Ahuserious's subscription feature PR) — subscription / vault-change-emitter
- Companion write-provenance and concurrency-improvement working drafts (relevant content summarized
  inline below — this doc is self-contained)
- **Already prototyped** (unmerged upstream, since this doc was first drafted): the audit-log
  side of "where provenance lives" (see open question 2, and the recommendation below) is built
  and sitting on branches, not just proposed —
  [`turbovault:feat/audit-metadata-passthrough`](https://github.com/ForrestThump/turbovault/tree/feat/audit-metadata-passthrough)
  (`*_with_metadata` `VaultManager` APIs — `write_file`/`edit_file`/`delete_file`/`move_file` can
  now attach arbitrary JSON, e.g. `WriteProvenance`, to the resulting `AuditEntry.metadata`) and
  [`turbovault:feat/write-note-request-meta`](https://github.com/ForrestThump/turbovault/tree/feat/write-note-request-meta)
  (threads an MCP `write_note` request's `_meta` — where a provenance field used on a local project lives — onto that
  same field automatically). Both are merged into `develop` on the `ForrestThump` fork. They
  depend on
  [`turbomcp:feature/tool-call-request-meta`](https://github.com/ForrestThump/turbomcp/tree/feature/tool-call-request-meta)
  (threads `_meta` through `tools/call` at the protocol level) — not yet published to crates.io,
  so the turbovault branches only build against that fork, not a released turbomcp.
- Issue [#32](https://github.com/Epistates/turbovault/issues/32) Git native write substrate.

---

## TL;DR

I want consumers of the vault event stream (PR #24's `subscribe_vault_events` /
`fetch_vault_events`) to be able to answer **"who caused this change?"** — primarily so reactive
agents can ignore changes their own writes produced and avoid infinite reaction loops.

The event stream is built on the **filesystem watcher**, so it is provenance-blind by construction:
a `FileModified` event is identical whether it came from a Turbovault write, from Obsidian, or from
`git`. Provenance is known only by the *writer*, never by the *event*. Bridging that gap is a
**correlation problem**, and there are two viable shapes:

[...]

Issue #29, opened 2026-06-06
Vector search as an optional feature-gated vertical


nicholasjpaterno:
## Summary

TurboVault's built-in semantic search (TF-IDF) is a strong low-resource default and will stay the default. This issue tracks adding **true embedding-based semantic search** as an **optional, feature-gated vertical** for larger or more semantically diverse vaults, where dense embeddings meaningfully improve recall over lexical/TF-IDF approaches.

This is not a replacement for TF-IDF. It is an opt-in upgrade, off by default, so the common case (modest vault, zero-config, instant, offline, private) is unaffected.

## Prior art

@ForrestThump has a working prototype on his `feature/vector-db` branch (CPU-friendly embeddings, paragraph-level incremental re-embedding, gated behind a `vector-search` feature) — excellent reference material. To be clear, this is open for anyone to pick up; no assumption about who authors it, and ForrestThump is very welcome to drive it if he has the bandwidth.

## Architecture direction (from a Q2 2026 review)

Design it as a vertical that abstracts the embedder (and optional reranker) behind traits, defaults to a pure-Rust CPU stack, and gates heavier backends behind features.

**Tier 1 — default backend (pure-Rust, CPU, edge-friendly)**
- Static embeddings via `model2vec-rs` + `hnsw_rs` (pure Rust, zero FFI; cross-compiles cleanly to Pi/musl/WASM).
- Model: Potion `potion-mxbai-256d-v2` (~71.45 MTEB v2 English, 15k+ sentences/sec on CPU, 256-dim so the index stays small); `potion-mxbai-micro` (~0.7 MB) for the edge floor.
- Static embeddings have closed most of the gap with small transformers while being orders of magnitude faster and smaller — a good fit for the local-first ethos.

**Tier 2 — optional quality upgrade (heavier dependency)**
- Quantized transformers via `fastembed-rs` (ONNX, int8): Qwen3-Embedding-0.6B (Matryoshka 256–2048 dims, Apache-2.0) or Nomic Embed v2 (CPU-purpose-built, single-model dense+sparse hybrid).

**Tier 3 — optional reranker (highest quality per cost)**
- FlashRank-class on CPU (sub-20 ms for ~50 candidates), or mxbai-rerank-xsmall / jina-reranker-v3 for higher quality.

**Hybrid note:** TurboVault already has Tantivy BM25. The strongest retrieval path is BM25 + dense + RRF, then optional rerank — mostly assembled from parts that already exist.

## Suggested shape
[...]

PR #39, opened 2026-07-19
feat: establish Phase 2 plugin API boundary

nicholasjpaterno:
## Summary

- add the publishable turbovault-plugin-api contract crate
- add a curated CAS-only VaultApi facade with no raw manager or server access
- adapt object-safe plugin providers into TurboMCP CompositeHandler mounts
- preserve flat core names while enforcing plugin_id_local_tool namespaces
- add a default-off plugin-api feature and future feature-wiring convention
- add a bounded best-effort hook bus with sequence, lag/resync, closure, content identity, and advisory provenance semantics
- document that plugin-owned direct Rust APIs are allowed and reviewed independently of the host boundary

## Safety and compatibility

- plugin writes must choose create-only or match a backend-native version token; blind overwrite is not exposed
- request context excludes raw transport sessions, principals, headers, and server internals
- descriptor, namespace, local tool, duplicate route, and duplicate prefix validation happens before serving
- the existing 74-tool core catalog remains byte-for-byte unchanged when no plugins are registered
- plugin-api is default-off, so current default binaries do not take on the dependency or surface

## Verification

- cargo test --workspace --all-targets --all-features --locked
- cargo clippy --workspace --all-targets --all-features -- -D warnings
- RUSTDOCFLAGS=-D warnings cargo doc --workspace --no-deps --all-features
- cargo package -p turbovault-plugin-api --locked --allow-dirty

The new API crate packages and registry-verifies. Registry verification of the parent turbovault package is expected to wait until the new dependency is published first; the documented release publish order now does that.

Advances Phase 2 of #34. Establishes the compatibility target for #19, #24, and #29, and carries the best-effort attribution contract discussed in #33. This intentionally does not close the umbrella issue.

### DIFF
--- a/crates/turbovault-plugin-api/src/hooks.rs
+++ b/crates/turbovault-plugin-api/src/hooks.rs
@@ -1,0 +1,326 @@
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
+use std::sync::{Arc, Mutex};
+use std::time::{SystemTime, UNIX_EPOCH};
+
+use serde::{Deserialize, Serialize};
+use tokio::sync::broadcast;
+
+/// Provenance supplied by a writer for best-effort event correlation.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct WriteProvenance {
+    /// Stable caller-selected source, such as `daily-review-agent`.
+    pub source: String,
+    /// Optional identifier connecting related operations.
+    pub correlation_id: Option<String>,
+    /// Optional human-readable reason.
+    pub note: Option<String>,
+}
+
+/// Best-effort event attribution.
+///
+/// This is advisory loop-prevention metadata, never an authorization or
+/// security boundary.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case", tag = "kind", content = "provenance")]
+pub enum EventAttribution {
+    /// The host correlated the observed content with a known write.
+    Attributed(WriteProvenance),
+    /// No trusted correlation was available; the change may be external.
+    ExternalOrUnknown,
+}
+
+/// A vault mutation observed by a hook producer.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case", tag = "kind")]
+pub enum HookEvent {
+    /// A note was created.
+    FileCreated {
+        /// Vault-relative path.
+        path: String,
+    },
+    /// A note was modified.
+    FileModified {
+        /// Vault-relative path.
+        path: String,
+    },
+    /// A note was deleted.
+    FileDeleted {
+        /// Vault-relative path.
+        path: String,
+    },
+    /// A note moved between two paths.
+    FileRenamed {
+        /// Original vault-relative path.
+        from: String,
+        /// New vault-relative path.
+        to: String,
+    },
+    /// The authoritative state must be re-read after a producer reset.
+    ResyncRequired {
+        /// Human-readable reason for the reset.
+        reason: String,
+    },
+}
+
+/// Sequenced event delivered to hook subscribers.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct VaultEventEnvelope {
+    /// Process-local monotonic sequence number.
+    pub sequence: u64,
+    /// Observation time as Unix epoch milliseconds.
+    pub observed_at_ms: u64,
+    /// Vault name.
+    pub vault: String,
+    /// Observed mutation.
+    pub event: HookEvent,
+    /// Content identity when it was available at observation time.
+    pub content_hash: Option<String>,
+    /// Best-effort writer attribution.
+    pub attribution: EventAttribution,
+}
+
+/// Current event-bus lifecycle.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum HookLifecycle {
+    /// New events may be published.
+    Running,
+    /// No further events will be accepted.
+    Closed,
+}
+
+/// Publish failure.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, thiserror::Error)]
+pub enum PublishError {
+    /// The bus was explicitly closed.
+    #[error("hook bus is closed")]
+    Closed,
+}
+
+/// Receive failure with explicit overflow semantics.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, thiserror::Error)]
+pub enum HookRecvError {
+    /// No event is currently available to a non-blocking receiver.
+    #[error("no hook event is currently available")]
+    Empty,
+    /// This subscriber fell behind the bounded ring buffer.
+    ///
+    /// The consumer must re-read authoritative vault state before continuing.
+    #[error("hook subscriber lagged by {skipped} event(s); resync required")]
+    Lagged {
+        /// Number of discarded envelopes.
+        skipped: u64,
+    },
+    /// The bus closed and all buffered events were drained.
+    #[error("hook bus is closed")]
+    Closed,
+}
+
+struct HookBusInner {
+    sender: Mutex<Option<broadcast::Sender<VaultEventEnvelope>>>,
+    sequence: AtomicU64,
+    closed: AtomicBool,
+    capacity: usize,
+}
+
+/// Cloneable bounded event bus shared by a host and its plugins.
+///
+/// Delivery is best-effort. A slow subscriber receives
+/// [`HookRecvError::Lagged`] and must resynchronize from [`crate::VaultApi`].
+#[derive(Clone)]
+pub struct HookBus {
+    inner: Arc<HookBusInner>,
+}
+
+impl std::fmt::Debug for HookBus {
+    fn fmt(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        formatter
+            .debug_struct("HookBus")
+            .field("capacity", &self.capacity())
+            .field("lifecycle", &self.lifecycle())
+            .finish()
+    }
+}
+
+impl HookBus {
+    fn sender(&self) -> std::sync::MutexGuard<'_, Option<broadcast::Sender<VaultEventEnvelope>>> {
+        self.inner
+            .sender
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner)
+    }
+
+    /// Create a bus with a fixed per-subscriber ring-buffer capacity.
+    ///
+    /// A zero capacity is promoted to one.
+    pub fn new(capacity: usize) -> Self {
+        let capacity = capacity.max(1);
+        let (sender, _) = broadcast::channel(capacity);
+        Self {
+            inner: Arc::new(HookBusInner {
+                sender: Mutex::new(Some(sender)),
+                sequence: AtomicU64::new(0),
+                closed: AtomicBool::new(false),
+                capacity,
+            }),
+        }
+    }
+
+    /// Return the fixed ring-buffer capacity.
+    pub fn capacity(&self) -> usize {
+        self.inner.capacity
+    }
+
+    /// Return whether new events are accepted.
+    pub fn lifecycle(&self) -> HookLifecycle {
+        if self.inner.closed.load(Ordering::Acquire) {
+            HookLifecycle::Closed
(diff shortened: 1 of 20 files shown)


docs/development/plugins.md
```diff
@@ -0,0 +1,79 @@
+# Plugin Architecture
+
+TurboVault's v1 plugin model is compiled-in, Cargo-feature-gated Rust. It is a
+curated extension surface: plugins are reviewed and shipped with TurboVault.
```
ForrestThump, line 4: "...shipped with TurboVault as optional Cargo features." Adds clarity

docs/development/plugins.md
```diff
@@ -0,0 +1,79 @@
...
+the Git backend.
+
+Plugins receive a curated request context containing request, user, session,
+client, and serializable metadata fields. Raw transport sessions,
+authentication principals, vault managers, and the MCP server never cross the
+boundary.
+
+## Registration and names
+
+The main crate's `plugin-api` feature is default-off. A vertical plugin feature
```
ForrestThump, line 29: Solid architecture idea.

docs/development/plugins.md
```diff
@@ -0,0 +1,79 @@
...
+plugin-tasks = ["plugin-api", "dep:turbovault-plugin-tasks"]
+```
+
+At startup, compiled-in factories are passed to
+`ObsidianMcpServer::new_with_plugins`. The host validates descriptors, rejects
+duplicate namespaces and tools, and wraps the object-safe provider for
+TurboMCP's clone-based composite handler.
+
+Core tool names stay flat for compatibility. Plugin tools are always
+advertised as `<plugin_id>_<local_tool>`. A `list_tasks` tool owned by the
```
ForrestThump, line 43: Does the default MCP prompt change depending on whether plugins are enabled to make this clear to agents? If plugins are disabled, it wouldn't make sense to explain the tool name format to an agent, but if they are enabled, then it might make sense to put this schema format into context. Setting it based on whether they are enabled could be nice for convenience.

docs/development/plugins.md
```diff
@@ -0,0 +1,79 @@
...
+shared envelope; external edits remain attribution-blind unless they can be
+correlated by content identity.
+
+## Plugin-owned Rust APIs
+
+The host contract governs only the capabilities used to mount a plugin into
+TurboVault. A plugin may also expose a broader Rust library API for direct
+non-MCP consumers. That public surface remains owned by the plugin maintainer
+and is reviewed for stability, dependency cost, and maintenance burden like
+any other in-tree public API; it does not need to be forced through `VaultApi`.
```
ForrestThump, line 69: I personally appreciate this design decision a lot. It helps me to maintain loose coupling and modularity in my projects. I acknowledge the additional review burden that it imposes on TurboVault maintenance.

docs/development/plugins.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Contribution rules
+
+- Put verticals under `crates/plugins/`.
+- Keep every vertical feature default-off.
+- Depend on `turbovault-plugin-api`; do not reach into server internals.
+- Publish only local tool names and let the host add the namespace.
+- Treat hook delivery and provenance as advisory.
+- Add provider-contract, namespace, lifecycle, lag/resync, and feature-off
+  tests.
```
ForrestThump, line 79: No rule for also having feature-on tests? Such a requirement could help make auditing and maintaining a high quality plugin suite easier to do. At the same time, obviously the plugins are default off, so a plugin that adheres to the existing rules and doesn't have robust internal testing would just be additional maintenance on the plugin maintainer, not on you. I guess it's a toss up. Interested in other opinions though.

docs/crates/README.md
```diff
@@ -15,6 +15,7 @@ The `TurboVault` project is organized as a Rust workspace with 8 crates:
 | **turbovault-batch** | Validated fail-fast operation batches | ~650 LOC | Conflict validation, sequencing, detailed results |
 | **turbovault-export** | JSON/CSV export formats | ~600 LOC | Data export, reporting |
 | **turbovault-tools** | Reusable MCP operations | ~2,500 LOC | Search, graph, metadata, and file operations |
+| **turbovault-plugin-api** | Stable plugin boundary | Small | Curated vault facade, provider contract, bounded hooks |
 | **turbovault-server** | CLI + MCP server | ~600 LOC | Main binary, server orchestration |
 
 **Total**: ~11,000 lines of production Rust code
```
ForrestThump, line 21: We are at 51,964 lines or Rust as of this PR, commit d58ac762dc749815215f363c5b63c6a02db2e2ef

crates/turbovault-plugin-api/src/vault.rs
```diff
@@ -0,0 +1,126 @@
...
+    host: Arc<dyn VaultHost>,
+}
+
+impl std::fmt::Debug for VaultApi {
+    fn fmt(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        formatter.debug_struct("VaultApi").finish_non_exhaustive()
+    }
+}
+
+impl VaultApi {
```
ForrestThump, line 101: I can't think of anything I'd change to this API.

crates/turbovault-plugin-api/src/vault.rs
```diff
@@ -0,0 +1,126 @@
...
+    pub path: String,
+    /// Complete replacement content.
+    pub content: String,
+    /// Mandatory create-or-CAS condition; blind overwrites are not exposed.
+    pub precondition: WritePrecondition,
+    /// Optional Git commit subject. Hosts may require it by policy.
+    pub commit_message: Option<String>,
+    /// Optional best-effort writer identity for hook correlation.
+    ///
+    /// This metadata is advisory and is not an authorization boundary.
```
ForrestThump, line 53: Good comment to have.

crates/turbovault-plugin-api/src/provider.rs
```diff
@@ -0,0 +1,140 @@
...
+    pub name: String,
+    /// Plugin version.
+    pub version: String,
+    /// Short human-readable purpose.
+    pub description: String,
+}
+
+impl PluginDescriptor {
+    /// Validate the descriptor before the host mounts any capabilities.
+    pub fn validate(&self) -> PluginResult<()> {
```
ForrestThump, line 25: Does this need to be any more robust to enforce MCP spec? Honest question, not sure.

crates/turbovault-plugin-api/src/hooks.rs
```diff
@@ -0,0 +1,326 @@
...
+            HookLifecycle::Running
+        }
+    }
+
+    /// Subscribe to future events.
+    pub fn subscribe(&self) -> Result<HookSubscription, PublishError> {
+        let guard = self.sender();
+        let sender = guard.as_ref().ok_or(PublishError::Closed)?;
+        Ok(HookSubscription {
+            receiver: sender.subscribe(),
```
ForrestThump, line 187: Verified that [subscribe() does not return a Result](https://docs.rs/tokio/latest/tokio/sync/broadcast/struct.Sender.html#method.subscribe).

crates/turbovault-plugin-api/src/hooks.rs
```diff
@@ -0,0 +1,326 @@
...
+    pub fn try_recv(&mut self) -> Result<VaultEventEnvelope, HookRecvError> {
+        self.receiver.try_recv().map_err(|error| match error {
+            broadcast::error::TryRecvError::Closed => HookRecvError::Closed,
+            broadcast::error::TryRecvError::Lagged(skipped) => HookRecvError::Lagged { skipped },
+            broadcast::error::TryRecvError::Empty => HookRecvError::Empty,
+        })
+    }
+}
+
+#[cfg(test)]
```
ForrestThump, line 263: Consider adding one or more comments to make it clear that HookBus::new()'s usize arg is its capacity. Just makes it a little easier to maintain / review. Can also make type clear via 

```
let bus = HookBus::new(2_usize);
```

Intellisense ghost annotates everything for humans though, and type declaration just duplicates in context for agents since they're going to look at the constructor though. Hmm I hadn't thought about the downsides of type annotation until now, but that is a real downside technically speaking. Probably leave that `2_usize` thing off. I'm leaving that in comment though for discussion / interest.

crates/turbovault/src/tools/plugin_host.rs
```diff
@@ -0,0 +1,199 @@
...
+            .map_err(map_host_error)?;
+        Ok(NoteSnapshot {
+            vault,
+            path: path.to_string(),
+            content,
+            version,
+        })
+    }
+
+    async fn write_note(&self, request: WriteNoteRequest) -> PluginResult<WriteReceipt> {
```
ForrestThump, line 111: Consider decomposing into one or more CoreToolHandler methods. It smells like future duplication.

crates/turbovault/src/tools/providers.rs
```diff
@@ -46,6 +52,127 @@ use self::templates::TemplateProvider;
...
+            let valid_name = !tool.name.is_empty()
+                && tool.name.chars().all(|character| {
+                    character.is_ascii_alphanumeric() || character == '_' || character == '-'
+                });
+            if !valid_name {
+                return Err(anyhow!(
+                    "plugin {:?} has invalid local tool name {:?}",
+                    descriptor.id,
+                    tool.name
+                ));
```
ForrestThump, line 81: I would 1) Extract this into a helper 2) Verify that it meets MCP specs 3) Put a comment on the helper that states that it enforces MCP spec for the tool name.

ForrestThump (commented):
Thanks for the API!

nicholasjpaterno:
@ForrestThump Apologies, I missed these comments! Looking them over now

docs/development/plugins.md
nicholasjpaterno: Thanks, that is clearer. I updated the wording in #40 to say that plugins ship with TurboVault as optional Cargo features.

docs/development/plugins.md
nicholasjpaterno: Good question. Currently in TurboMCP 3.1.5, it does not let McpHandler set InitializeResult.instructions. In #40 I used serverInfo.description instead. It only mentions the naming convention when plugins are actually registered. The default server stays unchanged, and tools/list remains the source of truth.

docs/development/plugins.md
nicholasjpaterno: I agree. Thanks for calling this out. #40 now requires tests with plugins enabled as well as compatibility tests with them disabled. Each plugin also needs focused tests in its own crate.

docs/crates/README.md
nicholasjpaterno: Thanks for checking the count. I removed the LOC estimates in #40 instead of replacing them with another number that would drift again.

docs/development/plugins.md
nicholasjpaterno: I'm currently in the process of writing TurboMCP v4 and i'll make sure it directly addresses this.

crates/turbovault-plugin-api/src/provider.rs
nicholasjpaterno: This was a good question. I checked the final MCP rule in SEP 986. #40 now has one public helper that enforces the allowed characters and the 64 character limit. It validates both the local name and the final namespaced name.

crates/turbovault-plugin-api/src/hooks.rs
nicholasjpaterno: Good maintenance point. In #40 the tests use named CAPACITY constants, with a short comment explaining why one test keeps two events and the other intentionally causes lag.

crates/turbovault/src/tools/plugin_host.rs
nicholasjpaterno: Good catch. I pulled the shared setup and cleanup into CoreToolHandler in #40. The core file provider and plugin host now use the same active vault lookup, backend selection, commit message policy, and cache invalidation path.

crates/turbovault/src/tools/providers.rs
nicholasjpaterno: Thanks, this was a useful nudge. #40 adds reusable validation helpers with a comment linking them to MCP SEP 986. The tests cover the allowed separators, invalid characters, empty names, the 64 character boundary, and names that become too long after namespacing.

crates/turbovault-plugin-api/src/vault.rs
ForrestThump: Just needed to sleep on it! #41

PR #41, opened 2026-07-19
feat(plugin-api): read-only `read_config` for module self-tuning to app settings

ForrestThump:
## What

Adds one read-only method to the curated plugin surface:

```rust
// VaultHost (default returns None), with a VaultApi passthrough
async fn read_config(&self, relative_path: &str) -> PluginResult<Option<Vec<u8>>>;
```

It lets a compiled-in module read a vault **application-config** file — e.g.
`.obsidian/plugins/obsidian-tasks-plugin/data.json` — so it can tune itself to
the user's app/plugin settings instead of duplicating them into module config.

## Why

The note APIs deliberately exclude dotfolders: `list_notes` returns vault-relative
paths (no filesystem anchor) and the scanner skips `.obsidian` and non-note
extensions. So there is currently **no way for a module to read app config**.
Without this, every module has to re-ask the user for settings the app already
stores.

## Design (kept deliberately minimal)

- **Name-addressed, not enumerable.** A module reads a config path it already
  knows (every plugin knows its own settings path). There is intentionally **no
  `list_configs`** — enumeration would be an authority amplifier (discover then
  read other plugins' stored secrets), and no consumer needs it. Keeping it
  name-addressed makes the capability capability-safe: you can only read a path
  you can already name.
- **`.obsidian/`-scoped + traversal-guarded** in `PluginVaultHost`, reusing the
  vault manager's existing `resolve_path` path-traversal detection.
- **`Ok(None)` on absent**, not `Err` — a missing file is a normal outcome (the
  target plugin may not be installed); the caller decides whether that is
  recoverable or fatal.
- **Non-breaking**: the trait method has a default returning `None`, so a host
  that doesn't support config reads degrades gracefully rather than erroring.

## Tests

- A `ContractProvider` contract test seeds an `.obsidian/` file and asserts it
  reads back by path, separately from the note space.
- **Live-validated against a real Obsidian vault.** The consuming `tasks` module
  (see below) has a fail-closed, env-gated live test that reads a real 2648-byte
  Tasks `data.json` and self-tunes to it — resolving `taskFormat` → emoji, the
  `#task` global filter, and matching 61 real tasks, with `source ==
  "obsidian-data"` (i.e. it genuinely read the file, not a heuristic fallback).
  Full real-vault scan in ~0.5s. Test:
  [`live_tasks_against_real_obsidian_vault`](https://github.com/ForrestThump/turbovault/blob/feat/plugin-tasks/crates/turbovault/src/tools/providers.rs).

## Context / consumer

This is the enabling capability for a `tasks` module that self-tunes to a user's
Obsidian Tasks settings (metadata format: emoji vs dataview; global filter),
rather than requiring separate configuration. The full module is not part of this
PR — it's here for reference and as end-to-end proof of the capability:
**[ForrestThump/turbovault @ `feat/plugin-tasks`](https://github.com/ForrestThump/turbovault/tree/feat/plugin-tasks)**.

I saw `fix/plugin-api-review-followups` is in flight (SEP-986 tool-name
validation, etc.) — happy to rebase/align this with wherever the API review
lands, or fold it into that work if you'd prefer.

### DIFF
--- a/crates/turbovault-plugin-api/src/vault.rs
+++ b/crates/turbovault-plugin-api/src/vault.rs
@@ -84,6 +84,23 @@ pub trait VaultHost: Send + Sync {
 
     /// Create or compare-and-swap a complete note.
     async fn write_note(&self, request: WriteNoteRequest) -> PluginResult<WriteReceipt>;
+
+    /// Read a read-only vault application-config file, e.g. an entry under
+    /// `.obsidian/`, returning `None` when it does not exist.
+    ///
+    /// This is the only door onto the vault's non-note config space (the note
+    /// APIs deliberately exclude dotfolders like `.obsidian`). It exists so a
+    /// module can self-tune to the user's app settings — for example the
+    /// Obsidian Tasks plugin's `data.json` — instead of requiring the settings
+    /// to be duplicated into module config.
+    ///
+    /// Hosts enforce read scoping and path-traversal safety, and MAY decline the
+    /// capability entirely; the default implementation returns `None` so that a
+    /// host which does not support config reads degrades gracefully rather than
+    /// erroring. The path is vault-relative and uses `/` separators.
+    async fn read_config(&self, _relative_path: &str) -> PluginResult<Option<Vec<u8>>> {
+        Ok(None)
+    }
 }
 
 /// Cloneable, curated facade supplied to every plugin.
@@ -122,5 +139,11 @@ impl VaultApi {
     /// Create or compare-and-swap a complete note.
     pub async fn write_note(&self, request: WriteNoteRequest) -> PluginResult<WriteReceipt> {
         self.host.write_note(request).await
+    }
+
+    /// Read a read-only vault application-config file (e.g. under `.obsidian/`),
+    /// returning `None` when it does not exist or the host declines the read.
+    pub async fn read_config(&self, relative_path: &str) -> PluginResult<Option<Vec<u8>>> {
+        self.host.read_config(relative_path).await
     }
 }
--- a/crates/turbovault/src/tools/plugin_host.rs
+++ b/crates/turbovault/src/tools/plugin_host.rs
@@ -192,6 +192,34 @@ impl VaultHost for PluginVaultHost {
             bytes: request.content.len(),
         })
     }
+
+    async fn read_config(&self, relative_path: &str) -> PluginResult<Option<Vec<u8>>> {
+        // Scope this capability to the Obsidian app-config namespace. Modules may
+        // read `.obsidian/**` (their settings live there); everything else stays
+        // behind the note APIs.
+        let normalized = relative_path.replace('\\', "/");
+        if normalized != ".obsidian" && !normalized.starts_with(".obsidian/") {
+            return Err(PluginError::invalid_input(format!(
+                "read_config is scoped to `.obsidian/`; refused {relative_path:?}"
+            )));
+        }
+
+        let manager = self
+            .core
+            .get_active_vault_manager()
+            .await
+            .map_err(map_host_error)?;
+        // resolve_path applies path_trav traversal detection against the vault root.
+        let resolved = manager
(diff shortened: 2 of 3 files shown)


nicholasjpaterno (changes requested):
Thanks for putting this together, ForrestThump. I like the direction and would like to get this landed.

Could we tighten the read boundary before merging? Right now every plugin gets the same `VaultApi`, and `read_config` can read any named file under `.obsidian`. Not listing paths helps, but a plugin can still read another plugin's known settings path.

What do you think about constructing `VaultApi` per plugin and giving each plugin an explicit list of allowed config paths, empty by default. A default `Plugin` method returning exact paths should be enough for Tasks and will also set us up cleanly for #42.

Could we also add tests for a missing file, the default host declining the read, a path outside `.obsidian`, `..` traversal, and an existing symlink escape on Unix? Please add the new method to the plugin guide and changelog too. The current `Ok(None)` behavior for a missing file looks right.

#40 has merged, so this will also need a rebase onto `main` and a fresh CI run.

Thanks again. Once those changes are in, this should be ready.

ForrestThump:
Yeah, I'll expand the test coverage. For the allowed paths, that definitely makes sense to make it more controlled. Would the idea be that a given plugin has a static &str for it's associated Obsidian plugin's folder that is passed to the VaultApi constructor, and then it can append subfiles and sub paths (with `../[some nefarious path here]` caught and rejected)? Definitely would be way easier to audit and maintain security over time. And if the static &str for the plugin folder path sounds right, would you want that to go in the plugin's crate? It could instead go in crates/turbovault-plugin-api/src/ in some file that had a list of the curated plugins... but... eh if the plugins are going to go in crates in this repo, that feels like it would just be a false sense of security through moving keys around. But it could potentially make sense if users were going to link crates from other repos against the api under features in the manifest. Thoughts?

ForrestThump:
Open question above. Still happy to work on this. In brief, "Where and how should the validated config paths be defined and stored?"

nicholasjpaterno:
Good call on both points, and sorry for leaving you hanging five weeks on a direct question. That one's on me.

Awkward bit: we ended up building this while you were waiting, so there's no commit in it for you. Not how I wanted it to go.

We went with the per-crate design you leaned toward, and your "false sense of security through moving keys around" line is what decided it. A central list in `turbovault-plugin-api` just duplicates what the plugin already knows and then drifts from it. It's a `Plugin::capabilities` method, granting nothing by default:

```rust
fn capabilities(&self) -> PluginCapabilities {
    PluginCapabilities::none()
        .with_config_read(".obsidian/plugins/obsidian-tasks-plugin/data.json")
}
```

Two things we added past your question. Exact file paths instead of a folder prefix, so you can see exactly what a plugin reads. And validation at mount time, so a bad declaration stops the server starting instead of turning into a confusing denial later. Traversal and symlink escapes get rejected, and declarations are normalized on the way in.

I'll close this when #56 lands, and the changelog credits #41. If you want to stay on plugin work, #42 and #43 are in that same PR, so #29 (vector search) is the open one, and your prototype already proved it out. Happy to hand it to you.

ForrestThump:
@nicholasjpaterno No worries! I'm sincerely very glad to hear that everything got shipped! Feel free to close this PR, and don't worry about credit stuff- I'm not very serious or sensitive when it comes to that. Yeah I'm happy to own the vector db plugin. I'll get something cooking on that based on the API updates, along with my task plugin.

Edit: Actually I closed it. Hope that's alright.

nicholasjpaterno:
Sorry for the long silence. Things got tied down on our end and your question sat a lot longer than it should have, especially with us shipping the thing you were asking about in the meantime. Appreciate you sticking around through that.

I've sent you triage invites In (turbovault and turbomcp) repos, should have done that a while back. No pressure but your persistent contributions are greatly appreciated @ForrestThump.

And #29 (vector search) is yours if you want it, since you already prototyped it. No deadline attached, just say the word.

PR #44, opened 2026-07-22
refactor(write-substrate): single mutation chokepoint via ChangePlan + WriteSubstrate

dlobue:
## Single write chokepoint: one `VaultManager.apply(ChangePlan)` for every mutation

This is the **write-surface** piece raised in #34 — now built, green, and merged
with the latest plugin-API feedback (#40). Read / graph / search were already
clean; **writes were the tangled part** — three parallel plan representations and
a per-call *"git or direct?"* dispatch smeared across the tool and server layers.
This PR collapses them into one primitive the Phase-2 `VaultApi` facade can sit on.

### At a glance

- **One door for every write.** Every mutation — single edit, batch, metadata,
  template, rollback-restore — funnels through `VaultManager`. No per-call
  "git or not?" branching above the manager.
- **One mutation vocabulary.** A backend-agnostic `ChangePlan` (in
  `turbovault-core`, git-free) describes any write as *ordered changes +
  per-path preconditions + a message*. Single write = 1-change plan; batch =
  N-change plan. Replaces **three** incompatible representations.
- **Two backends behind one `apply`.** `VaultManager` picks `direct`
  (filesystem) or `git` (commit-per-write) from config; each implements exactly
  `apply(&ChangePlan)`. Nothing above the manager — including the MCP server —
  knows which is live.
- **Domain tools own their operation again.** `MetadataTools` / `TemplateEngine`
  / `BatchTools` compute-and-write *through the manager*, not the
  split-compute-hoist-write shape the cutover scaffolding left behind.
- **One conflict error.** Every "changed underneath us" failure surfaces as a
  single `core::ConcurrencyError` on both backends.
- **Net effect: less surface.** The git substrate becomes an implementation
  detail *inside* the manager (a feature-gated backend) rather than a second
  write path beside it.

```mermaid
classDiagram
    class ChangePlan {
        <<turbovault-core, git-free>>
        message
        changes : Upsert / Remove / Rename
        preconditions : per-path
    }
    class VaultManager {
        <<single write chokepoint>>
        write_file / edit_file / delete_file / move_file
        apply_changes(plan)
    }
    class WriteSubstrate {
        <<enum, one per vault>>
        apply(plan)
    }
    class DirectSubstrate {
        filesystem write
        token = sha256
    }
    class GitSubstrate {
        one commit per plan
        token = blob-oid
    }
    class DomainTools {
        FileTools / MetadataTools
        TemplateEngine / BatchTools
    }
    DomainTools --> VaultManager : compute, then write
[...]

PR #45, opened 2026-07-27
feat(write-substrate): carry caller metadata from ChangePlan onto audit entries

ForrestThump:
## What

`ChangePlan` gains an optional `metadata: Option<serde_json::Value>`, and `DirectSubstrate`
records it verbatim on every audit entry the plan produces.

```rust
let plan = ChangePlan::new("write")
    .create("notes/n.md", body)
    .with_metadata(json!({ "source": "daily-review-agent", "correlation_id": "r-1" }));

// -> AuditEntry { path: "notes/n.md", metadata: {"source": "...", ...}, .. }
```

Absent metadata is unchanged behaviour: the entry keeps its default `{}`.

## Why

A filesystem change event is provenance-blind. A `FileModified` is identical whether Turbovault,
Obsidian, or `git` wrote the file — only the *writer* ever knows who it was, and there was nowhere
for a writer to say so.

That leaves a consumer of the vault unable to distinguish its own writes from a human's. The
concrete failure: an agent generates a note, files it into the vault, then observes that note
change, cannot attribute it, and reacts to its own output — repeatedly, until a depth limit stops
it.

`AuditEntry.metadata` already exists and is already `serde_json::Value`. Nothing could reach it.
This connects the two ends.

## Why here

Before #44 this had to be threaded tool by tool, and every new write path re-opened the hole.
#44 made `ChangePlan` the single mutation vocabulary and `DirectSubstrate::record_audit` the single
place audit entries are written — so this lands **once** and covers every entry point: the single-op
mutators, batch, and anything built later.

That is the whole reason this is a small diff.

## Design notes

- **Opaque `serde_json::Value`, not a typed provenance struct.** It lands verbatim in a field of the
  same type, so nothing here needs to know or agree on a schema, and `turbovault-core` takes no new
  dependency. Callers put back whatever they read: provenance, correlation ids, trace context. A
  typed accessor (`plugin-api`'s `WriteProvenance` serialises straight into this slot) can layer on
  top later without changing this plumbing.
- **Plan-scoped, not per-change.** One plan is one logical mutation, so its entries share one
  origin. A multi-change plan stamps all of them.
- **Turbovault does not interpret it.** It is carried, never read — advisory correlation data, not
  an authorization boundary.
- **`GitSubstrate` is untouched**, since it records no audit entries today. When it grows them, the
  field is already on the plan it receives.

## Tests

Four new tests in `substrate.rs`, using a substrate with audit + snapshot wired (the existing
`direct()` helper leaves both `None`, which makes `record_audit` a no-op):

- metadata reaches a create/update entry, with the rest of the entry unchanged
- metadata reaches `Move` and `Delete` entries
- every entry of a multi-change plan carries it
- a plan without metadata still records the default `{}`

`cargo fmt` clean; `clippy` adds no new warnings.

**Pre-existing failures, unrelated to this change:** on Windows, 4 tests in
`test_mcp_provider_workflows` fail identically with and without this patch
(`audit_create_delete_and_unsupported_rollbacks_...`, `templates_render_notes_...`,
`graph_navigation_and_topology_...`, `file_lifecycle_and_concurrency_guards_...`). They assert
forward-slash path suffixes and Unix `"No such file"` error strings. Verified by stashing the patch
and re-running against pristine `main`.

## Follow-up (not in this PR)

The MCP tool layer can forward a request's `_meta` into `ChangePlan::with_metadata`, which is what
makes this reachable by an MCP client rather than only an in-process caller. That is deliberately
excluded here: it needs a `turbomcp` that surfaces request `_meta` to handlers, which is not on the
released crate yet. This PR is useful on its own and unblocks that one.

Related: #33 (write provenance in the vault event stream).

### DIFF
--- a/crates/turbovault-core/src/change_plan.rs
+++ b/crates/turbovault-core/src/change_plan.rs
@@ -50,6 +50,25 @@ pub struct ChangePlan {
     pub message: String,
     pub changes: Vec<Change>,
     pub preconditions: Vec<(String, Precondition)>,
+    /// Opaque, caller-defined metadata recorded on the audit entries this plan
+    /// produces. **Turbovault does not interpret it.**
+    ///
+    /// It exists because a filesystem change event is provenance-blind: a
+    /// `FileModified` is identical whether Turbovault, Obsidian, or git wrote
+    /// the file. Only the *writer* knows who it was, and before this there was
+    /// nowhere for a writer to say so, so a reactive consumer could not tell
+    /// its own writes apart from a human's — and reacted to notes it had just
+    /// produced itself. Recording the caller's own metadata on the audit entry
+    /// lets a consumer join a change back to the write that caused it and
+    /// break that loop.
+    ///
+    /// Deliberately `serde_json::Value` rather than a typed provenance struct:
+    /// it lands verbatim in [`turbovault_audit::AuditEntry::metadata`], which
+    /// is already this type, so nothing here needs to know or agree on a
+    /// schema. Callers put whatever they read back — provenance, correlation
+    /// ids, trace context. A typed accessor can layer on top later without
+    /// changing this plumbing.
+    pub metadata: Option<serde_json::Value>,
 }
 
 impl ChangePlan {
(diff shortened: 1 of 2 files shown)


nicholasjpaterno:
Merging. Landing this at the chokepoint rather than threading it tool by tool is the whole reason it is a small diff, and you are right that `AuditEntry.metadata` was already the right shape with nothing able to reach it.

Keeping it an opaque `serde_json::Value` was the right restraint. A typed provenance struct would have forced everyone to agree on a schema before anyone could use one, and `turbovault-core` picks up no new dependency this way.

#47 landed just ahead of you and restructured the same apply loop, so I merged main into your branch and resolved it: its loop, your metadata. One consequence worth knowing, a partially applied plan now stamps metadata on the entries that did land, which is what lets a consumer attribute the prefix it can actually see. Verified locally, 1211 tests pass with clippy and fmt clean.

PR #46, opened 2026-07-30
feat(write-safety): clobber-safety matrix + first-class wire preconditions

dlobue:
# The Write-Safety Suite: pinning clobber-safety across every write layer

Builds on #44 (single mutation chokepoint). Adds the test suite that says what the
write surface MUST do, and closes the last gaps where the wire could not express it.

Here is the spreadsheet that started it all: https://docs.google.com/spreadsheets/d/11dpI_9YLIjDqfwZZJxKNoOauGpJ9rwO8g36atpMuywA/edit?usp=sharing

A csv export of that spreadsheet is included in the PR in docs/write-safety-suite/wss-precondition-matrix.csv along with the full design here: docs/write-safety-suite/README.md

## The promise this is about

> TurboVault's promise is to **ABORT** if the content on disk doesn't match what you
> said the file looks like.

Nothing verified that promise systematically. This adds a matrix that sets up **all
possible scenario combinations** and checks TurboVault aborts when it *should*.

Stated as a boundary: WSS tests **clobber-safety** (did the op refuse-or-proceed
without silently losing an out-of-band change?), **not content-correctness** (is the
written text formatted right?). Those are orthogonal; the second belongs in ordinary
functional tests and never in a WSS cell.

## What's in it

**1. The suite** — `crates/turbovault/tests/write_safety_suite/` + the
`wss_matrix` runner. Four axes, deliberately separated:

| Axis | What varies |
|---|---|
| **Backend** | `Git` (full 9-state working-tree grid) · `Direct` (`{absent, present}`) |
| **Layer (World)** | `ToolsWorld` · `ManagerWorld` · `BatchWorld` · `WireWorld` (in-process `ObsidianMcpServer`) |
| **Op** | write · edit · delete · update_frontmatter · manage_tags · create_from_template · move (src + dest roles) |
| **Cell** | `(precondition × working-tree state) → required outcome` |

The load-bearing rule: **every layer runs the op's ONE shared `Case` table.** A layer
that behaves differently *fails that cell* — the divergence is the finding. Forking a
per-layer table to record divergent behaviour would enshrine the bug, so it is
explicitly forbidden in the authoring rules.

**2. The wire precondition surface** — the gaps the suite exposed. Preconditions now
cross every wire/serde boundary as a sentinel-or-oid string
(`expected_hash: <40-hex-oid> | "absent" | "exists" | "blind"`):
- `update_frontmatter`, `manage_tags`, `create_from_template` **gain** `expected_hash`
  (they previously hardcoded their precondition and silently ignored the caller).
- `move_note` **gains** `dest_expected_hash` — a first-class destination precondition,
  replacing a hardcoded expect-absent. Same for `BatchOperation::MoveNote`.
- `write_note`'s `force` is **retired** into the sentinel scheme (see Breaking below).

**3. Error unification** — the git dirty-gate refusal now raises `ConcurrencyError`
instead of `ConfigError`. Same refusal, correct kind: "something changed underneath
us" is one condition, and callers were unable to distinguish it from a config problem.

## ⚠️ Breaking / wire-visible

- **`write_note` loses `force`.** `force: true` becomes `expected_hash: "blind"`. The
  create-on-existing refusal now surfaces as `ConcurrencyError` (was
  `invalid_request`), with an actionable message. Omitting the param preserves current
  behaviour, and the default is mode-dependent (overwrite → expect-absent,
  append/prepend → expect-exists).
- **`delete_note`'s `force` is untouched** — it conflates a backlink-bypass with a
  precondition and needs its own decision (tracked separately).
- Everything else is additive: omitting a new param reproduces today's behaviour. The
  golden `tool_catalog.json` is regenerated accordingly.

## Reviewing it

```bash
just wss-test      # the gate: active cells (pending are #[ignore]d)
just wss-report    # ONE table, emoji status + legend — paste-able into a comment
just wss-audit     # do the Rust tables still match the source-of-truth CSV?
```

`docs/write-safety-suite/README.md` is the spec; the CSVs beside it are authoritative,
and `just wss-audit` mechanically enforces that the code matches them (0 drift today:
253 spec cells ↔ 253 grid cells, no missing/extra/mismatched outcomes).

## Aspirational by design — please read this before the red

**The suite states required behaviour, not current behaviour.** 301 cells are marked
`pending` (ignored trials) because TurboVault does not satisfy them yet — that set IS
the burndown, and it is the point of shipping this now rather than after.

- `cargo test --test wss_matrix` → **791 active cells, all passing.** CI stays green.
- `--include-ignored` → the 301 pending cells, of which 296 currently fail.
- The reporter is the reconcile oracle: it lists **un-pend candidates** (pending cells
  that now pass → activate them) and **newly-failing** (a regression). Both lists are
  the review signal; newly-failing is 0.

The 5 current un-pend candidates are known *batch-ahead* cells — the batch path already
gives idempotent delete-of-absent and NoFile-before-precondition on move, while the
standalone ops do not. They stay pending until the standalone ops catch up, so the
matrix never claims a behaviour is universal when only one path has it.

## Known gaps (deliberate, documented)

- **`write_note` append/prepend** — not covered. An append never destroys bytes, so an
  append with a *stale* `ExpectBlob` is arguably not a clobber at all; the design says
  "in-place → expect-exists + dirty-gated" in one place and lists "append/prepend CAS
  semantics" as an open question in another. **This is a maintainer call** — writing
  either answer into the authoritative CSV would publish a contract nobody ratified.
  The layers also disagree on whether the mode exists (`FileTools` and the wire take a
  `WriteMode`; `VaultManager::write_file` and `BatchOperation` do not).
- **`BatchOperation::UpdateLinks`** — the only batch variant with no arm, pending a
  decision on whether it survives (it is a naive, non-OFM-aware `str::replace`).

Neither is a coverage oversight; `just wss-audit` proves the tables match the spec for
everything the spec covers.

## The suite tests itself

A matrix is only as trustworthy as its setup, so the harness carries its own probes:

- **Setup fidelity** — re-derives each cell's working-tree state from git
  *independently* of the code that built it. If `build_state(etcsu)` silently produced
  `etc-u`, every cell in that column would test the wrong scenario and still pass.
- **Precondition N/A rule** — a token that fails to resolve does not *fail* a cell, it
  silently *deletes* it. The probe asserts resolvability matches the rule, so a
  coverage hole can't masquerade as a clean run.
- **Plumbing fidelity** — a minimal op run on **all four** layers asserting the
  observations are *identical*. The matrix checks each layer separately against a
  coarse outcome, so two layers can diverge in ways no cell catches.
- **Backend identity** — a git-backed layer's successful write must advance `HEAD`.
  Without this, a layer wired to the direct substrate over a git repo would pass every
  `Ok` cell with the whole suite green against the wrong substrate.
- **Wire shape** — every param the wire layer sends is checked against the server's own
  tool catalog. The wire is the only layer whose command shape isn't compiler-checked;
  a typo'd param name is silently ignored, which makes a write succeed unconditionally.

Each was verified by fault injection (deliberately breaking the thing it guards and
confirming it fails loudly). The misspelled-param case ships as a `#[should_panic]`
test, so that guard re-proves itself on every run.

## Verification

- `just test` green (fmt + clippy `-D warnings` + `--workspace --all-features` + doctests).
- Matrix: 791 active passing / 0 newly-failing / 301 pending. `just wss-audit`: 0 drift.
- Mutation testing on the write path: 22 caught / 5 missed / 9 unviable; a
  workspace-scoped re-check reduced the survivors to 3, all the EXDEV cross-device
  rename fallback (killing it needs a rename spanning two filesystems inside one vault).
- Coverage on the code the suite exercises: `substrate.rs` 96.0%, `manager.rs` 97.4%,
  `precondition.rs` 100%. `cargo-crap`: `turbovault-vault` 112 functions, 0 crappy.
- dupehound slop on the suite: 1.0% (grade A).

## Note on the commit list

The branch predates #44 and carries its own copies of the relayering commits, which
were squash-merged upstream. The **diff is correct** — those files are identical to
`upstream/main` — but the commit list shows them. Happy to rebase onto `upstream/main`
to drop the duplicates if you'd prefer a clean history.


dlobue:
Here is what `just wss-report` outputs:

# WSS write-safety matrix

Legend — every cell is <status><outcome>: the outcome WSS requires, and whether the
code currently delivers it.

  status   ✅ active, passing — required behaviour holds
           ❌ active, FAILING — a regression; this list must be empty
           ⏳ pending, failing — known burndown, expected (turbovault-nbl.8)
           🔶 pending, PASSING — un-pend candidate: activate this cell
           · no test: not constructible (backend cannot build the state, or the
             precondition's token is undefined for it — the CSV's N/A)
           ⏩ no test: duplicate of another precondition whose token resolves
             equal in this state (the CSV's SKIP:X)

  outcome  Ok  succeeded and materialised
           CE  ConcurrencyError — refused with NO disk change (CAS mismatch or
               dirty-tree refusal; the matrix's DIRTY_ERR and CAS_FAIL unify here)
           NF  NoFile — in-place op on an absent target (FileNotFound)
           OE  OpError — op-specific refusal, no disk change (e.g. edit SEARCH
               matched nothing)

  precondition (checked against the WORKING TREE)
           BLIND    no precondition; last-writer-wins
           ABSENT   ExpectAbsent — create-only
           EXISTS   ExpectExists — must exist, any content (in-place default)
           HEAD     ExpectBlob(HEAD oid)    — defined iff committed
           INDEX    ExpectBlob(INDEX oid)   — defined iff staged
           WORKDIR  ExpectBlob(WORKDIR oid) — defined iff it exists
           WRONG    ExpectBlob(bogus oid)   — never matches

  state    [e]xists[t]racked[c]ommitted[s]taged[u]nstaged (columns, in CSV order)

Cells are addressable as trials: <world>::<backend>::<operation>::<PRECONDITION>::
<state>::<expected> — copy one into `cargo test --test wss_matrix -- <name>`.

| world   | backend | operation            | precondition | ----- | etc-- | etcs- | etc-u | etcsu | et-s- | et--u | et-su | e---u |
|---------|---------|----------------------|--------------|-------|-------|-------|-------|-------|-------|-------|-------|-------|
| tools   | git     | write_note           | BLIND        | ✅Ok  | ✅Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| tools   | git     | write_note           | ABSENT       | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| tools   | git     | write_note           | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| tools   | git     | write_note           | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| tools   | git     | write_note           | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| tools   | git     | write_note           | WRONG        | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| tools   | git     | edit_note            | EXISTS       | ✅NF  | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| tools   | git     | edit_note            | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| tools   | git     | edit_note            | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| tools   | git     | edit_note            | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| tools   | git     | edit_note            | WRONG        | ✅NF  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| tools   | git     | delete_note          | EXISTS       | ⏳Ok  | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| tools   | git     | delete_note          | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| tools   | git     | delete_note          | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| tools   | git     | delete_note          | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| tools   | git     | delete_note          | WRONG        | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| tools   | git     | update_frontmatter   | EXISTS       | ✅NF  | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| tools   | git     | update_frontmatter   | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| tools   | git     | update_frontmatter   | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| tools   | git     | update_frontmatter   | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| tools   | git     | update_frontmatter   | WRONG        | ✅NF  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| tools   | git     | manage_tags          | EXISTS       | ✅NF  | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| tools   | git     | manage_tags          | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| tools   | git     | manage_tags          | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| tools   | git     | manage_tags          | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| tools   | git     | manage_tags          | WRONG        | ✅NF  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| tools   | git     | create_from_template | BLIND        | ✅Ok  | ✅Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| tools   | git     | create_from_template | ABSENT       | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| tools   | git     | move_note::src       | EXISTS       | ⏳NF  | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| tools   | git     | move_note::src       | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| tools   | git     | move_note::src       | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| tools   | git     | move_note::src       | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| tools   | git     | move_note::src       | WRONG        | ⏳NF  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| tools   | git     | move_note::dest      | BLIND        | ✅Ok  | ✅Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| tools   | git     | move_note::dest      | ABSENT       | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| tools   | git     | move_note::dest      | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| tools   | git     | move_note::dest      | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| tools   | git     | move_note::dest      | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| tools   | git     | move_note::dest      | WRONG        | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| tools   | direct  | write_note           | BLIND        | ✅Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| tools   | direct  | write_note           | ABSENT       | ✅Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| tools   | direct  | write_note           | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| tools   | direct  | write_note           | WRONG        | ✅CE  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| tools   | direct  | edit_note            | EXISTS       | ✅NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ⏳CE  |
| tools   | direct  | edit_note            | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| tools   | direct  | edit_note            | WRONG        | ✅NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| tools   | direct  | delete_note          | EXISTS       | ⏳Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ⏳CE  |
| tools   | direct  | delete_note          | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| tools   | direct  | delete_note          | WRONG        | ✅CE  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| tools   | direct  | update_frontmatter   | EXISTS       | ✅NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ⏳CE  |
| tools   | direct  | update_frontmatter   | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| tools   | direct  | update_frontmatter   | WRONG        | ✅NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| tools   | direct  | manage_tags          | EXISTS       | ✅NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ⏳CE  |
| tools   | direct  | manage_tags          | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| tools   | direct  | manage_tags          | WRONG        | ✅NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| tools   | direct  | create_from_template | BLIND        | ✅Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| tools   | direct  | create_from_template | ABSENT       | ✅Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| tools   | direct  | move_note::src       | EXISTS       | ⏳NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ⏳CE  |
| tools   | direct  | move_note::src       | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| tools   | direct  | move_note::src       | WRONG        | ⏳NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| tools   | direct  | move_note::dest      | BLIND        | ✅Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| tools   | direct  | move_note::dest      | ABSENT       | ✅Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| tools   | direct  | move_note::dest      | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| tools   | direct  | move_note::dest      | WRONG        | ✅CE  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| manager | git     | write_note           | BLIND        | ✅Ok  | ✅Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| manager | git     | write_note           | ABSENT       | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| manager | git     | write_note           | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| manager | git     | write_note           | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| manager | git     | write_note           | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| manager | git     | write_note           | WRONG        | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| manager | git     | edit_note            | EXISTS       | ✅NF  | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| manager | git     | edit_note            | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| manager | git     | edit_note            | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| manager | git     | edit_note            | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| manager | git     | edit_note            | WRONG        | ✅NF  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| manager | git     | delete_note          | EXISTS       | ⏳Ok  | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| manager | git     | delete_note          | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| manager | git     | delete_note          | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| manager | git     | delete_note          | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| manager | git     | delete_note          | WRONG        | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| manager | git     | move_note::src       | EXISTS       | ⏳NF  | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| manager | git     | move_note::src       | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| manager | git     | move_note::src       | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| manager | git     | move_note::src       | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| manager | git     | move_note::src       | WRONG        | ⏳NF  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| manager | git     | move_note::dest      | BLIND        | ✅Ok  | ✅Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| manager | git     | move_note::dest      | ABSENT       | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| manager | git     | move_note::dest      | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| manager | git     | move_note::dest      | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| manager | git     | move_note::dest      | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| manager | git     | move_note::dest      | WRONG        | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| manager | direct  | write_note           | BLIND        | ✅Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| manager | direct  | write_note           | ABSENT       | ✅Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| manager | direct  | write_note           | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| manager | direct  | write_note           | WRONG        | ✅CE  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| manager | direct  | edit_note            | EXISTS       | ✅NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ⏳CE  |
| manager | direct  | edit_note            | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| manager | direct  | edit_note            | WRONG        | ✅NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| manager | direct  | delete_note          | EXISTS       | ⏳Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ⏳CE  |
| manager | direct  | delete_note          | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| manager | direct  | delete_note          | WRONG        | ✅CE  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| manager | direct  | move_note::src       | EXISTS       | ⏳NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ⏳CE  |
| manager | direct  | move_note::src       | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| manager | direct  | move_note::src       | WRONG        | ⏳NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| manager | direct  | move_note::dest      | BLIND        | ✅Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| manager | direct  | move_note::dest      | ABSENT       | ✅Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| manager | direct  | move_note::dest      | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| manager | direct  | move_note::dest      | WRONG        | ✅CE  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| batch   | git     | write_note           | BLIND        | ✅Ok  | ✅Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| batch   | git     | write_note           | ABSENT       | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| batch   | git     | write_note           | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| batch   | git     | write_note           | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| batch   | git     | write_note           | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| batch   | git     | write_note           | WRONG        | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| batch   | git     | edit_note            | EXISTS       | ✅NF  | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| batch   | git     | edit_note            | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| batch   | git     | edit_note            | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| batch   | git     | edit_note            | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| batch   | git     | edit_note            | WRONG        | ✅NF  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| batch   | git     | delete_note          | EXISTS       | 🔶Ok  | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| batch   | git     | delete_note          | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| batch   | git     | delete_note          | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| batch   | git     | delete_note          | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| batch   | git     | delete_note          | WRONG        | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| batch   | git     | update_frontmatter   | EXISTS       | ✅NF  | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| batch   | git     | update_frontmatter   | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| batch   | git     | update_frontmatter   | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| batch   | git     | update_frontmatter   | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| batch   | git     | update_frontmatter   | WRONG        | ✅NF  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| batch   | git     | manage_tags          | EXISTS       | ✅NF  | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| batch   | git     | manage_tags          | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| batch   | git     | manage_tags          | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| batch   | git     | manage_tags          | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| batch   | git     | manage_tags          | WRONG        | ✅NF  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| batch   | git     | create_from_template | BLIND        | ✅Ok  | ✅Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| batch   | git     | create_from_template | ABSENT       | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| batch   | git     | move_note::src       | EXISTS       | 🔶NF  | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| batch   | git     | move_note::src       | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| batch   | git     | move_note::src       | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| batch   | git     | move_note::src       | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| batch   | git     | move_note::src       | WRONG        | 🔶NF  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| batch   | git     | move_note::dest      | BLIND        | ✅Ok  | ✅Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| batch   | git     | move_note::dest      | ABSENT       | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| batch   | git     | move_note::dest      | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| batch   | git     | move_note::dest      | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| batch   | git     | move_note::dest      | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| batch   | git     | move_note::dest      | WRONG        | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| batch   | direct  | write_note           | BLIND        | ✅Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| batch   | direct  | write_note           | ABSENT       | ✅Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| batch   | direct  | write_note           | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| batch   | direct  | write_note           | WRONG        | ✅CE  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| batch   | direct  | edit_note            | EXISTS       | ✅NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ⏳CE  |
| batch   | direct  | edit_note            | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| batch   | direct  | edit_note            | WRONG        | ✅NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| batch   | direct  | delete_note          | EXISTS       | ⏳Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ⏳CE  |
| batch   | direct  | delete_note          | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| batch   | direct  | delete_note          | WRONG        | ✅CE  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| batch   | direct  | update_frontmatter   | EXISTS       | ✅NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ⏳CE  |
| batch   | direct  | update_frontmatter   | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| batch   | direct  | update_frontmatter   | WRONG        | ✅NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| batch   | direct  | manage_tags          | EXISTS       | ✅NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ⏳CE  |
| batch   | direct  | manage_tags          | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| batch   | direct  | manage_tags          | WRONG        | ✅NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| batch   | direct  | create_from_template | BLIND        | ✅Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| batch   | direct  | create_from_template | ABSENT       | ✅Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| batch   | direct  | move_note::src       | EXISTS       | 🔶NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ⏳CE  |
| batch   | direct  | move_note::src       | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| batch   | direct  | move_note::src       | WRONG        | 🔶NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| batch   | direct  | move_note::dest      | BLIND        | ✅Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| batch   | direct  | move_note::dest      | ABSENT       | ✅Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| batch   | direct  | move_note::dest      | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| batch   | direct  | move_note::dest      | WRONG        | ✅CE  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| wire    | git     | write_note           | BLIND        | ✅Ok  | ✅Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| wire    | git     | write_note           | ABSENT       | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| wire    | git     | write_note           | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| wire    | git     | write_note           | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| wire    | git     | write_note           | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| wire    | git     | write_note           | WRONG        | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| wire    | git     | edit_note            | EXISTS       | ✅NF  | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| wire    | git     | edit_note            | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| wire    | git     | edit_note            | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| wire    | git     | edit_note            | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| wire    | git     | edit_note            | WRONG        | ✅NF  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| wire    | git     | delete_note          | EXISTS       | ⏳Ok  | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| wire    | git     | delete_note          | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| wire    | git     | delete_note          | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| wire    | git     | delete_note          | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| wire    | git     | delete_note          | WRONG        | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| wire    | git     | update_frontmatter   | EXISTS       | ✅NF  | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| wire    | git     | update_frontmatter   | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| wire    | git     | update_frontmatter   | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| wire    | git     | update_frontmatter   | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| wire    | git     | update_frontmatter   | WRONG        | ✅NF  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| wire    | git     | manage_tags          | EXISTS       | ✅NF  | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| wire    | git     | manage_tags          | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| wire    | git     | manage_tags          | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| wire    | git     | manage_tags          | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| wire    | git     | manage_tags          | WRONG        | ✅NF  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| wire    | git     | create_from_template | BLIND        | ✅Ok  | ✅Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| wire    | git     | create_from_template | ABSENT       | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| wire    | git     | move_note::src       | EXISTS       | ⏳NF  | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| wire    | git     | move_note::src       | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| wire    | git     | move_note::src       | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| wire    | git     | move_note::src       | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| wire    | git     | move_note::src       | WRONG        | ⏳NF  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| wire    | git     | move_note::dest      | BLIND        | ✅Ok  | ✅Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| wire    | git     | move_note::dest      | ABSENT       | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| wire    | git     | move_note::dest      | HEAD         | ·     | ✅Ok  | ✅CE  | ✅CE  | ✅CE  | ·     | ·     | ·     | ·     |
| wire    | git     | move_note::dest      | INDEX        | ·     | ·     | ⏳Ok  | ·     | ✅CE  | ⏳Ok  | ·     | ✅CE  | ·     |
| wire    | git     | move_note::dest      | WORKDIR      | ·     | ⏩    | ⏩    | ⏳Ok  | ⏳Ok  | ⏩    | ⏳Ok  | ⏳Ok  | ⏳Ok  |
| wire    | git     | move_note::dest      | WRONG        | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  | ✅CE  |
| wire    | direct  | write_note           | BLIND        | ✅Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| wire    | direct  | write_note           | ABSENT       | ✅Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| wire    | direct  | write_note           | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| wire    | direct  | write_note           | WRONG        | ✅CE  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| wire    | direct  | edit_note            | EXISTS       | ✅NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ⏳CE  |
| wire    | direct  | edit_note            | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| wire    | direct  | edit_note            | WRONG        | ✅NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| wire    | direct  | delete_note          | EXISTS       | ⏳Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ⏳CE  |
| wire    | direct  | delete_note          | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| wire    | direct  | delete_note          | WRONG        | ✅CE  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| wire    | direct  | update_frontmatter   | EXISTS       | ✅NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ⏳CE  |
| wire    | direct  | update_frontmatter   | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| wire    | direct  | update_frontmatter   | WRONG        | ✅NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| wire    | direct  | manage_tags          | EXISTS       | ✅NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ⏳CE  |
| wire    | direct  | manage_tags          | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| wire    | direct  | manage_tags          | WRONG        | ✅NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| wire    | direct  | create_from_template | BLIND        | ✅Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| wire    | direct  | create_from_template | ABSENT       | ✅Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| wire    | direct  | move_note::src       | EXISTS       | ⏳NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ⏳CE  |
| wire    | direct  | move_note::src       | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| wire    | direct  | move_note::src       | WRONG        | ⏳NF  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| wire    | direct  | move_note::dest      | BLIND        | ✅Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| wire    | direct  | move_note::dest      | ABSENT       | ✅Ok  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |
| wire    | direct  | move_note::dest      | WORKDIR      | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅Ok  |
| wire    | direct  | move_note::dest      | WRONG        | ✅CE  | ·     | ·     | ·     | ·     | ·     | ·     | ·     | ✅CE  |

## Summary

| operation | pass | fail | pending |
|---|---:|---:|---:|
| write_note | 132 | 56 | 56 |
| edit_note | 114 | 32 | 32 |
| delete_note | 105 | 39 | 40 |
| update_frontmatter | 84 | 24 | 24 |
| manage_tags | 84 | 24 | 24 |
| create_from_template | 45 | 21 | 21 |
| move_note | 232 | 100 | 104 |
| **TOTAL** | **796** | **296** | **301** |

## Un-pend candidates (5)

_Pending cells that PASS — activate them (`pending` -> `new`)._

- 🔶 `batch::direct::move_note::src::EXISTS::-----::NoFile`
- 🔶 `batch::direct::move_note::src::WRONG::-----::NoFile`
- 🔶 `batch::git::delete_note::EXISTS::-----::Ok`
- 🔶 `batch::git::move_note::src::EXISTS::-----::NoFile`
- 🔶 `batch::git::move_note::src::WRONG::-----::NoFile`

## Newly-failing (0)

_None._

crates/turbovault/src/tools/providers/metadata.rs
```diff
@@ -251,7 +263,13 @@ impl MetadataProvider {
             .resolve_commit_message(commit_message, || format!("move_file {from} -> {to}"))
             .await?;
         FileTools::new(manager)
-            .move_file_with_hash(&from, &to, expected_hash.as_deref(), &message)
+            .move_file(
+                &from,
+                &to,
+                turbovault_core::Precondition::for_in_place(expected_hash.as_deref()),
+                turbovault_core::Precondition::Blind,
```
dlobue, line 270: Hard-coding this precondition is not ideal... This is a gap in the WSS now that i think about it... the providers/vaultapi need a layer in WSS to ensure they live up to the same no-clobber promise.

crates/turbovault/src/tools/providers/metadata.rs
```diff
@@ -144,6 +153,7 @@ impl MetadataProvider {
         path: String,
         operation: String,
         tags: Option<Vec<String>>,
+        expected_hash: Option<String>,
```
dlobue, line 156: Ideally everything, including the MCP API, would talk in preconditions. I didn't feel comfortable making a sweeping change like that without talking it through with everyone else. At the moment, we overloaded `expected_hash` to accept sentinel values like `absent`, `exists`, or `blind` in addition to hashes and oids.

crates/turbovault-tools/src/file_tools.rs
```diff
@@ -144,7 +143,7 @@ impl FileTools {
             path,
             content,
             WriteMode::Overwrite,
-            None,
+            Precondition::Blind,
```
dlobue, line 146: This is another big gap. These backwards compatible helper "method-zoos" as claude likes to call them bypass the guarantees made by WSS. Collapsing all the different `write_file`/`write_file_with_mode`/`create_file` methods into a single `write_file` is another call I didn't feel comfortable making on my own, and wanted @nicholasjpaterno 's call on it.

crates/turbovault-tools/src/file_tools.rs
dlobue, line 235: huh, a tool that somehow didn't get swept up and put through the single mutation manager... grrr...

dlobue:
## ⚠️ Please hold off merging — the matrix in this PR asserts a contract we've since reversed

Flagging this before it gets merged, because the problem is in the part of the PR that
describes itself as authoritative.

**As published, `wss-precondition-matrix.csv` says `Blind × every dirty state → Ok`** —
i.e. a write carrying no precondition may overwrite a file with uncommitted changes. That
has now been reconsidered and settled the other way.

### The ratified policy

A write to a git-backed path with uncommitted changes is **refused**. `Blind` remains the
explicit force hatch, and is **disabled by default**. Both are per-vault config, and the
suite will verify both settings.

This is the behaviour the code already has, so the default is not a breaking change.

The reasoning, all three alternatives, and why two of them were rejected are now written
up in-repo at `docs/write-safety-suite/dirty-tree-policy.md`. Short version: committing,
stashing, or capturing the displaced work all preserve *bytes*, but none preserve the
signal `git add` carries — which is simultaneously "keep this" and "**not yet**". Those
readings point opposite ways, which is the strongest argument that an automated system
shouldn't be adjudicating between them. Refusal is the only option that declines to decide,
and the only one whose failure mode is recoverable.

A consequence worth stating plainly: under strict, a dirty path never reaches the CAS check,
so most of what the pending set describes as "behaviour to implement" turns out to be
"the code was already right and the spec was wrong."

### Also landing before this is mergeable

- **The CSV is now keyed by `backend`.** git and direct genuinely disagree, because direct
  is git-blind — its version token is the sha256 of the bytes, so staged/committed are
  invisible to it. 18 direct cells asserting `ConcurrencyError` on a merely-present file
  were corrected to `Ok`; they pass immediately with zero production changes, which is the
  evidence the requirement was wrong rather than the code.
- **`wss-audit` now checks both backend arms**, and reports specified-but-not-constructed
  cells as a counted total rather than skipping them silently, so a coverage gap can't read
  as coverage.

### Responding to your two review comments — both confirmed

**`providers/metadata.rs` (hard-coded precondition / no WSS layer for the providers):**
confirmed and worse than it looks. The wire arm covers seven tools; `move_file` also mutates,
is agent-reachable, hardcodes `Blind` on its *destination*, and has **zero cells**. Beyond
adding it, the fix wants a meta-test that enumerates mutating tools from the server's own
catalog and fails when one lacks an invoker — otherwise coverage drifts silently the next
time a tool is added.

**`file_tools.rs` (a tool that escaped the mutation manager):** confirmed — `copy_file` uses
raw `tokio::fs::copy`, so no precondition, no audit, no snapshot, no reindex, and on a git
vault no commit. It violates R1 directly. Worth noting it was an explicit M4 deliverable
("FileTools absorbs `copy_file` … routed through the manager here") that was marked done and
didn't land — the un-done half of a change is invisible in a diff, which is precisely how it
survived. Severity is capped by reach: it's SDK surface, not an MCP tool.

One further finding while tracing that: **`materialize` ends with `index.read_tree(&tree)`**,
which git2 documents as replacing the *entire* index. That's why the staged-changes check has
to be repo-wide — and it means narrowing that check (an obvious-looking tidy-up, since a vault
is currently unwritable whenever anything anywhere is staged) would silently convert a
usability wart into data loss. git2 already exposes per-path index updates; the two have to
move together.

I'd convert this to draft but don't have the permission from here — treating this comment as
the equivalent signal.

nicholasjpaterno:
Sorry this sat so long. Two things that were blocking it are now cleared on my side.

The Security Audit failure was never yours. `main` was failing `cargo audit` on four advisories published after your branch point, and clippy 0.1.98 also turned `drain_collect` into a hard error. #54 fixed both, so a rebase should turn that check green without you touching anything.

Main has also moved a fair amount underneath you: #45, #47, #48, #49, #50 and #53 all landed today. #47 is the one to look at, since it restructured `DirectSubstrate::apply` and gave `ApplyOutcome` a `failed_at` and an `error`, which is squarely in your area and may let you delete some of what this PR carries.

Could you rebase onto main and take it out of draft when you have time? The matrix is something I want, particularly the part that separates clobber-safety from content-correctness. That distinction is the reason a suite like this stays useful instead of rotting into assertions nobody trusts. No rush, just tell me if you would rather I take it from here.

nicholasjpaterno:
Good news on the rebase: I merged your branch onto current main locally and it's one trivial conflict, in `batch_tools.rs`, where your `apply_dest_precondition` and main's `op_records`/`changed_paths` landed at the same spot. Unrelated functions, keep both. One test also needs `dest_expected_hash: None` added at `test_batch_tools.rs:671`.

The `wss_matrix` suite is great, by the way. 809 passing in 16 seconds.

One real problem though. With your branch merged, `core_mcp_writes_reach_plugin_subscribers` in `test_plugin_boundary.rs` hangs, so a core `write_note` isn't reaching the hook bus (or deadlocks before it can). That test is green on main, so something in the merged write path drops the publish. Everything else passes.

That test hung instead of failing, which is our bug, and #61 fixes it so you'll get a five second failure with a message rather than a wedged run. Merge main after that lands and you should see it straight away.

No idea yet which of your changes does it. Happy to dig in if you'd rather I take that part.

nicholasjpaterno:
Found the hang, and it isn't your code. The boundary test's overwrite step still passes `force: true`, which this branch removes in favour of `expected_hash: "blind"`. So the write is correctly refused, nothing gets published, and the test waits for a modify event that can never arrive. One line in `test_plugin_boundary.rs` and it's green.

I pushed the whole thing to `wss-merged-probe` on the main repo if you want to diff against it. Current main merged in, the `batch_tools.rs` conflict resolved by keeping both functions, `dest_expected_hash: None` added at `test_batch_tools.rs:671`, and the tool catalog re-blessed with `UPDATE_TOOL_CATALOG=1`. 2132 passing, clippy and fmt clean.

The catalog diff is exactly what you'd want it to be: `force` gone from `write_note`, `dest_expected_hash` added to move, still 74 tools. An old client sending `force: true` gets a clear `is_error` back telling it to pass `expected_hash` or `"blind"`, so nothing silently no-ops. That does make it a wire break though, so I'm going to land this as 3.0.0 rather than a minor.

Thanks for your patience, this one sat much longer than it should have. Take it out of draft whenever you're ready.

nicholasjpaterno:
@dlobue I've opened #84 with this merged onto current `main` so it doesn't keep drifting. Your commits are in there as-is, and it's green. It'll go out as 3.0.0 because of the `force` removal. A review from you on #84 would be great if you have time, and `copy_file` is still open as a follow-up if you want it. I'll close this one once #84 lands.

PR #47, opened 2026-08-01
fix(batch): report what a direct batch actually did

dlobue:
On the `direct` backend, a partially-applied batch reports `executed: 0`, `failed_at: null`
and empty `changes`/`records` — while the earlier operation's file is genuinely on disk.
Found by dogfooding. Both the legacy executor and the git backend report this accurately, so
it is a regression rather than a long-standing gap.

## Root cause

`DirectSubstrate::apply` has two phases. The precondition gate checks every precondition
before applying anything and aborts the whole plan with nothing written — genuinely atomic.
The apply loop then used `?`, so a mid-loop failure **discarded the accounting for changes
already written** and returned `Err`. `BatchTools::batch_execute` funnels any `Err` into an
envelope hardcoded to `executed: 0`.

`ApplyOutcome.atomic` and `.failed_at` were likewise hardcoded to `true`/`None`. That was
accurate when only single-change plans existed; it became false once batches were routed
through the same path, and the fields were already carried to the wire for exactly this.

## What changes

- A **gate** failure still returns `Err` — nothing was written, the caller retries the whole plan.
- An **apply-loop** failure returns `Ok(ApplyOutcome)` carrying what landed, `failed_at`, and the
  error. `VaultManager` indexes the prefix that landed, so a partially-applied plan is no longer
  invisible to the link graph, file cache and search index.
- `batch_execute` translates that into per-operation reporting. Because the substrate counts
  *changes* and the wire reports *operations*, the plan builder now also returns per-op change
  spans — which is what lets `records[].affected_files` name backlink-rewritten linkers without
  a per-op branch.

## Wire / API notes

`ApplyOutcome` gains `error: Option<Error>` and consequently **loses `Clone`**, since `Error`
wraps `io::Error`. Anything calling `.clone()` on it breaks. `Arc<Error>` would restore it if
you would rather keep the derive — happy to switch.

## It corrects four assertions currently on `main`

#44 changed these from correct values to `executed: 0`, each with a comment deferring the real
values to a later phase:

- `test_tool_visibility_integration.rs::batch_response_reports_partial_failure_contract`
  (`executed == 1`, `failed_at == 1` → `0`/`null`)
- `test_async_error_paths.rs::test_batch_tools_partial_failure_rollback` (`2` → `0`)
- `test_batch_tools.rs::test_batch_execute_rollback_on_error` (`1` → `0`)
- `test_batch_tools.rs::test_batch_execute_atomic_guarantees` (`1` → `0`)

Each asserts a few lines later that the earlier file *does* exist, so `executed: 0` contradicted
the same test's own next assertion.

## On the test scenario

The partial is induced by an I/O failure (an upsert whose parent component is a regular file),
deliberately **not** by deleting an absent file — `wss-precondition-matrix.csv` specifies
delete-of-absent as `Ok`, so a test built on that would go red the moment the direct backend
honours it. The four tests above all still use delete-of-absent and will need the same treatment
when that cell lands; tracked separately.

## Overlap

Touches the same functions as #45 (caller metadata onto audit entries). Whichever lands first,
happy to rebase.

### DIFF
--- a/crates/turbovault-core/src/change_plan.rs
+++ b/crates/turbovault-core/src/change_plan.rs
@@ -39,6 +39,17 @@ impl Change {
             Change::Rename { from, .. } => from,
         }
     }
+
+    /// Every path this change mutates: `path` for `Upsert`/`Remove`, BOTH
+    /// endpoints for `Rename`. [`ChangePlan::touched_paths`] is the
+    /// plan-wide fold of this; a caller holding only a SUB-RANGE of a plan's
+    /// changes (the batch's per-operation span) folds it itself.
+    pub fn touched_paths(&self) -> Vec<&str> {
+        match self {
+            Change::Upsert { path, .. } | Change::Remove { path } => vec![path],
+            Change::Rename { from, to } => vec![from, to],
+        }
+    }
 }
 
 /// A backend-agnostic description of one mutation: an ordered set of
@@ -187,17 +198,11 @@ impl ChangePlan {
     /// `commit_changeset`) and by the tool layer's intra-batch path-collision
     /// check.
     pub fn touched_paths(&self) -> Vec<String> {
-        let mut out = Vec::with_capacity(self.changes.len());
-        for c in &self.changes {
-            match c {
-                Change::Upsert { path, .. } | Change::Remove { path } => out.push(path.clone()),
-                Change::Rename { from, to } => {
-                    out.push(from.clone());
-                    out.push(to.clone());
-                }
-            }
-        }
-        out
+        self.changes
+            .iter()
+            .flat_map(Change::touched_paths)
+            .map(String::from)
+            .collect()
     }
 }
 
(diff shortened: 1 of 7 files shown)


nicholasjpaterno:
Merging. This is a real regression and the diagnosis is exactly right: the gate is genuinely atomic, the apply loop was not, and `?` threw away the accounting for changes that were already on disk. Reporting `executed: 0` while the first file sits there is the worst version of that, because it reads as "nothing happened" to anyone deciding whether to retry.

Restoring the four assertions #44 weakened is the part I am glad you did rather than left. Those comments deferring real values to a later phase were a promise, and this is the phase.

On `ApplyOutcome` losing `Clone`: leave it. Nothing in tree needs it and `Arc<Error>` would be papering over the fact that an outcome carrying a live `io::Error` is not a value you should be copying around.

Heads up that #45 lands next and touches the same loop, adding plan metadata to the `upsert`/`remove`/`rename` calls. They compose fine, I have resolved it on my side, so no action needed from you.

PR #48, opened 2026-08-01
test(graph): regression cover for batch/move link-index reindex

dlobue:
A dogfooding report said the link graph was left stale by `batch_execute` CreateNote and by
`move_note` — links present on disk, `get_forward_links` empty until the file was touched by a
single `write_note`.

**It does not reproduce on current `main`.** The single-mutation-chokepoint work closed it:
`sync_index` re-parses every changed markdown path and calls `graph.add_file` +
`graph.update_links`; `apply_changes` calls `sync_index` for every plan; and both
`batch_execute` and the link-aware move route through there. One funnel, both paths.

This PR adds the regression cover so it stays closed. No production change.

## What is asserted

Six tests, three scenarios across both backends:

- a note created via `batch_execute` has its outgoing wikilinks in the graph
- `move_note` reindexes both endpoints
- a batch-created linker is visible to a later move

Every assertion runs with **no intervening `write_note`** — that touch was the workaround the
original report described, so admitting one would defeat the test.

## The tests can fail

Green tests prove nothing unless they can go red, so the graph-update block in `sync_index` was
disabled and all six failed with the original report's exact symptom (`expected a link ending in
"target.md", got []`), then reverted.

### DIFF
--- a/crates/turbovault/tests/test_graph_reindex_write_paths.rs
+++ b/crates/turbovault/tests/test_graph_reindex_write_paths.rs
@@ -1,0 +1,295 @@
+//! TV-015 (turbovault-8v8) regression cover: the link graph must be current
+//! the moment a write returns, on EVERY write path that changes link
+//! structure — not just single `write_note`.
+//!
+//! The original dogfood report (filed against the pre-relayering build) had
+//! two manifestations:
+//!
+//! 1. `batch_execute` `CreateNote` of a note carrying wikilinks left the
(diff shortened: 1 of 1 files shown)


nicholasjpaterno:
Merging this. The part that makes it worth having is the falsifiability check: disabling the graph-update block in `sync_index` and confirming all six go red with the original symptom, rather than trusting green tests on a path that was already fixed.

Thanks for chasing a report that turned out not to reproduce and leaving cover behind so it stays that way.

PR #49, opened 2026-08-01
refactor(server): one activation path for publishing a VaultManager

dlobue:
A git-backend `VaultManager` could be inserted into the server's cache **before** its reindex
machinery was wired. Once that happened, `get_active_vault_manager` cache-hit the unwired
manager forever and search staleness never closed.

That was fixed by wiring-before-publish at all three insert sites — which made the invariant
hold by discipline. A fourth insert site can still forget.

## What changes

The three open-coded build/wire/publish sequences collapse into one private
`activate_vault_manager(name) -> Arc<VaultManager>` that does build → audit → initialize → wire →
cache-insert as a single operation and hands back only the published `Arc`. A caller can no
longer publish an unwired manager because it never holds one.

Net **-31 lines**. Chosen over a `VaultManager::activate` constructor because the audit state
this path attaches (`AuditLog`/`SnapshotStore`, plus the server's own maps) lives above the vault
crate — pushing it down would have been a larger diff for the same guarantee.

This also resolves the `initialize_audit_for_manager` wrinkle that forced the multi-step publish:
it still needs `&mut` between `new()` and the `Arc`, but exactly one function spans that window
and it is the function that publishes.

## Two intended consequences

- **`--init` now attaches audit state.** `initialize_registered_vaults` previously skipped it, so
  `get_audit_tools` failed on any `--init`'d vault — the same species of forgotten wiring. Fixed
  by unification.
- **`add_vault` publishes first-writer-wins** instead of overwriting. Two callers activating one
  name concurrently would previously each build *and wire* a manager and then both insert,
  leaving two managers for one vault each running its own reindex tasks. The loser's manager is
  now dropped and its `Drop` aborts the tasks it started.

### DIFF
--- a/crates/turbovault/src/tools.rs
+++ b/crates/turbovault/src/tools.rs
@@ -370,29 +370,9 @@ impl CoreToolHandler {
                     continue;
                 }
             }
-            let vault_config = self
-                .multi_vault_mgr
-                .get_vault_config(&name)
+            self.activate_vault_manager(&name)
                 .await
-                .map_err(|e| Error::config_error(format!("get_vault_config({name}): {e}")))?;
-            let mut server_config = ServerConfig::default();
-            let mut vault_cfg = vault_config;
-            vault_cfg.is_default = true;
-            server_config.vaults = vec![vault_cfg];
-            let manager = VaultManager::new(server_config)
-                .map_err(|e| Error::config_error(format!("VaultManager::new({name}): {e}")))?;
-            manager
-                .initialize()
-                .await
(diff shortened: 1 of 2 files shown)


nicholasjpaterno:
Merging. Making the invariant structural rather than a discipline is the right call here, a fourth insert site cannot reintroduce the bug because it never holds an unwired manager.

Both of the consequences you flagged were real: `--init` silently skipping audit state broke `get_audit_tools` on those vaults, and the concurrent-activation path really did leave two managers per vault each running its own reindex tasks. Good catch on the second one, that would have been miserable to diagnose later.

PR #50, opened 2026-08-01
feat(vault): choose the write backend when registering a vault

dlobue:
`add_vault`, `create_vault` and the `--vault` CLI shorthand all build a `VaultConfig` without
setting `write_backend`, and the builder defaults to `Direct`. So **every runtime-registered vault
is a direct vault** — over MCP an agent cannot create or attach a git-backed one at all. The only
routes today are a startup `--config` file or the SDK, neither of which is wire-exposed.

## What changes

An optional `write_backend` plus a `backend_opts` object on both MCP tools and the CLI shorthand,
threaded into the builder. **Omitting them reproduces today's behaviour exactly**, and the YAML
config format is untouched.

On naming: the options argument is `backend_opts`, not `git`, because `write_backend` selects the
backend and its companion argument should not be named after one particular backend — if `direct`
ever gains settings there is already somewhere to put them. The config file keeps its `git:` key;
that wire/config asymmetry is deliberate rather than an oversight, and the alternative (moving
either backend's options in the YAML) would break existing configs for cosmetics.

Passing `backend_opts` together with the direct backend is an **error naming both values**, not a
silently dropped parameter.

## Also: the inverse footgun

Registering a path that *is* a git repository as a `Direct` vault means every write lands on the
filesystem and never commits — the working tree drifts from HEAD with no signal. This adds a
warning at registration.

Warning only. Refusing, or auto-detecting and switching the backend, are both wire-visible
behaviour changes that look like yours to make; a warning adds the signal without changing
anything that currently works.

## Note

`VaultLifecycleTools::build_config` became `pub` so the CLI shorthand and the MCP tools share one
guard rather than keeping two copies of the builder-plus-warning block. Additive.

### DIFF
--- a/crates/turbovault-core/src/config.rs
+++ b/crates/turbovault-core/src/config.rs
@@ -31,6 +31,24 @@ pub enum WriteBackend {
     Git,
 }
 
+/// turbovault-kdq: parse the backend out of a runtime registration parameter
+/// (the `write_backend` MCP argument, the `--vault-write-backend` CLI flag).
+/// Accepts exactly the serde spellings, `legacy` alias included, so a config
+/// value and a wire value can never disagree.
+impl std::str::FromStr for WriteBackend {
+    type Err = Error;
+
+    fn from_str(value: &str) -> Result<Self> {
+        match value.trim().to_ascii_lowercase().as_str() {
+            "direct" | "legacy" => Ok(Self::Direct),
+            "git" => Ok(Self::Git),
+            other => Err(Error::config_error(format!(
+                "Unknown write_backend '{other}' (expected 'direct' or 'git')"
+            ))),
+        }
+    }
+}
+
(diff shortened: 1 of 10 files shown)


nicholasjpaterno:
Merging. I had written a narrower version of this myself and yours is better, so I am dropping mine: covering `create_vault` as well as `add_vault`, funnelling both through `build_config` so they cannot drift, and refusing `backend_opts` with `direct` instead of quietly dropping them are all things mine did not do. The direct-over-git-repo warning is a good catch too, silent drift from HEAD is exactly the kind of thing nobody notices until it matters.

One gap left, and it was the one piece mine had: registering `write_backend=git` against a path that is not a repository still succeeds and then fails at the first write, well after the caller could connect the error to the registration call. `VaultRepo::open` at registration turns that into an immediate, actionable error. I will add it as a follow-up rather than hold this.

PR #53, opened 2026-08-25
feat(partial-reads): line and section slices on read_note

### LINKED ISSUE #52: Feature Request: Partial Reads for `read_note` (Line Slices and Heading Sections)
## Description

`read_note` always returns the complete content of a note. For large notes this
is wasteful for the common case where a caller needs only a small, well-defined
part of the file — the last few entries of an append-only log, or a single dated
section of a long-running note.

This request proposes two additive, optional selectors on `read_note`:

1. **Positional** — return the first or last _N_ lines.
2. **Structural** — return the first or last _N_ heading-delimited sections,
   optionally filtered by heading level and a pattern on the heading text.

Both share one response contract: a verbatim substring of the file plus explicit
metadata saying what was omitted. The default behaviour isn't changed.

## Use Case

Append-only notes like progress logs keep growing while the useful information is limited to the tail of the note. A project progress log using one `##` section per entry:

```markdown
## 2026-08-20
Did: ...
Decided: ...
Next: ...

## 2026-08-21
...
```

An agent resuming work needs the **last few entries**. Today it must read the
whole note — for a note with hundreds of entries, the overwhelming majority of
the returned content is discarded immediately. The cost is paid in the caller's
context budget.

Related shapes with the same problem:

- Daily notes and meeting logs where only recent entries matter.
- Long reference notes where a caller wants one named section.
- Any note used as an audit or changelog trail.

The positional selector covers the simpler case: tailing a note whose structure
is unknown or irregular, or sampling the head of an unfamiliar file to decide
whether to read it fully.

## Current Behavior

`read_note` (`crates/turbovault/src/tools/providers/files.rs`) reads the entire
file and returns `path`, `content`, `hash`, and `uri`. The call chain is:

```
read_note(path)                          providers/files.rs
  FileTools::read_file(&path)             turbovault-tools/src/file_tools.rs
    VaultManager::read_file(&path)        turbovault-vault/src/manager.rs
      tokio::fs::read_to_string(...)
```

There is no way to request a subset. No tool in the current catalogue exposes
line ranges, pagination, or section selection, so there is no established
convention to follow. This request also proposes the convention.

Note that `VaultManager::read_file` deliberately bypasses the parsed
`VaultFile` cache in order to return raw bytes including frontmatter. Any
partial-read feature must preserve that guarantee: what is returned should be a
verbatim substring of the file, never content reconstructed from a parsed representation.

## Proposed Solution

Add an optional selector to `read_note`. When no selector is supplied, behaviour
is byte-for-byte unchanged.

### Implementation Details

1. **New primitive in `turbovault-tools`**

   Add to `FileTools` in `file_tools.rs`, alongside the existing `read_file`:

   ```rust
   pub async fn read_file_slice(&self, path: &str, slice: Slice) -> Result<SliceResult>
   ```

   `read_file` stays untouched. The new function reads the file via the same
   `VaultManager::read_file` call and then slices the resulting `String`, so the
   raw-content guarantee is inherited rather than reimplemented. Slicing after a
   full read keeps this change small. A streaming read that stops early is a
   possible follow-up, discussed under Open Questions.

2. **Positional selector**

   `head_lines: Option<usize>` and `tail_lines: Option<usize>`. Lines rather
   than bytes or characters: lines match how callers reason about notes and
   cannot split a UTF-8 codepoint. Slicing is done on line boundaries and the
   returned text is joined verbatim.

3. **Structural selector via existing parser API**

   Use `turbovault_parser::parse_headings(content) -> Vec<Heading>`. Each
   `Heading` carries a `SourcePosition` with `line`, `column`, `offset`, and
   `length`, so heading positions are byte offsets into the content that was
   passed in. This allows verbatim slicing between offsets.

   This is a new *consumer* of an existing public API. No change to
   `turbovault-parser` is required.

   Using the parser rather than a line scan matters for correctness, because `parse_headings` does not treat comment lines inside fenced code blocks
   as headings. 

   Note that `ContentBlock::Heading` (the block AST in `turbovault-core`) carries
   no line or offset information — only `ContentBlock::Code` has
   `start_line`/`end_line`. The `parse_headings` API is the correct entry point
   for this feature, not the full block AST.

   Parameters:

   - `heading_level: Option<u8>` — which level delimits a section (e.g. `2`).
   - `heading_pattern: Option<String>` — match against heading text; selects
     only sections whose heading matches.
   - `last_sections: Option<usize>` / `first_sections: Option<usize>` — take
     from the end or start of the filtered list.

4. **Section boundary rule**

   A section runs from its heading's `offset` to the offset of the next heading of the **same or higher level**, or end of file. Not simply the next heading — otherwise a `##` section would be truncated at its own `###` subheading. This rule is the main thing worth covering with tests.

5. **Response contract**

   On a partial read, the response gains explicit metadata:

   - `truncated: true`
   - `total_lines` — line count of the complete file
   - `returned_lines: [start, end]` — 1-indexed, inclusive
   - `sections_matched` / `sections_returned` — structural selector only
   - `hash` — **omitted** (see below)

6. **Omit the hash on partial reads**

   `read_note` currently returns a `hash` that acts as a version token for
   optimistic concurrency: callers pass it back as `write_note`'s
   `expected_hash`, which becomes `Precondition::ExpectBlob` and causes the
   write to be refused if the file changed in the meantime.

   Returning a whole-file hash alongside a partial read would be unsafe. The
   token asserts "the caller has seen the current state of this file", which is
   false. A caller could read 50 lines, submit those 50 lines as a whole-file
   write with a valid `expected_hash`, and the precondition would pass and silently
   destroy the rest of the note.

   Omitting the hash makes `write_note` fall back to
   `Precondition::for_replace(None, false)` → `ExpectAbsent`, which refuses to
   clobber an existing file. Mutation paths remain available and are unaffected:

   - `edit_note` — targeted in-place change; `for_in_place(None)` →
     `ExpectExists`, needs no whole-file token. This is the intended follow-up.
   - `write_note` after a fresh full `read_note` — normal hashed write.
   - `write_note` with `force: true` — explicit, caller-acknowledged
     last-writer-wins. An escape hatch, not a recommended path.

   This reasoning applies to both selectors and applies more strongly to the
   structural one.

7. **Tool metadata**

   The `#[tool]` attribute on `read_note` should be updated so callers learn the
   right follow-up without reasoning about preconditions: extend `usage` to note
   that partial reads omit the hash, add `edit_note` to `related`, and have
   `next_steps` in the truncated response point at `edit_note`.

8. **No facade or schema changes required**

   `read_note` stays `read_only = true`; its name and route are unchanged, so
   `providers.rs` and its tool-count assertion need no edits. The JSON schema
   advertised to clients is derived from the function signature by the `#[tool]`
   macro and regenerates automatically.

## Examples

Last three entries of a progress log:

```json
{ "path": "projects/log.md", "heading_level": 2, "last_sections": 3 }
```

Dated entries only, ignoring other `##` sections:

```json
{ "path": "projects/log.md", "heading_level": 2,
  "heading_pattern": "^\\d{4}-\\d{2}-\\d{2}$", "last_sections": 3 }
```

Tail of a note with irregular structure:

```json
{ "path": "inbox/scratch.md", "tail_lines": 40 }
```

Response shape for a partial read:

```json
{
  "path": "projects/log.md",
  "content": "## 2026-08-20\nDid: ...\n\n## 2026-08-21\n...",
  "truncated": true,
  "total_lines": 1284,
  "returned_lines": [1201, 1284],
  "sections_matched": 142,
  "sections_returned": 3,
  "uri": "obsidian://..."
}
```

Unchanged default — no selector, full content, hash present:

```json
{ "path": "projects/log.md" }
```

Offsets from `parse_headings` confirmed against a log-shaped input:

```
input: "# Log\n\nintro\n\n## 2026-08-20\nDid: a\n\n## 2026-08-21\nDid: b\n..."

L1  lvl1 off=0  len=6  "Log"
L5  lvl2 off=14 len=14 "2026-08-20"
L8  lvl2 off=36 len=14 "2026-08-21"
L11 lvl2 off=58 len=14 "2026-08-22"
```

`&content[36..58]` is the `2026-08-21` section, verbatim. Offsets remain
absolute when frontmatter is present, so they align with the raw content
`read_note` already returns.

## Benefits

1. **Context efficiency**: Cost of reading a log becomes proportional to the
   request rather than to the note's history.
2. **Verbatim content**: Slicing by parser offsets returns exact bytes; nothing
   is reconstructed or normalised.
3. **Correct heading detection**: Reuses `parse_headings`, so fenced code blocks
   containing `#` are not mistaken for headings.
4. **No new safety surface**: Omitting the version token on partial reads keeps
   the existing optimistic-concurrency guarantees intact.
5. **Establishes a convention**: Gives future tools a partial-read shape to
   follow.

## Impact

- **Non-breaking**: All parameters optional; omitting them preserves current
  behaviour, including the returned `hash`.
- **Additive**: One new function in `FileTools`; `read_file` untouched.
- **Scope**: Two files — `crates/turbovault-tools/src/file_tools.rs` (primitive)
  and `crates/turbovault/src/tools/providers/files.rs` (tool signature,
  metadata, response). No change to `turbovault-parser`, `turbovault-vault`,
  `turbovault-core`, or `providers.rs`.
- **Tool count unchanged**: A parameter addition, not a new tool, so the
  `tool_routes` assertion in `providers.rs` is unaffected.

## Dependencies

- No new external dependencies.
- Uses the existing public `turbovault_parser::parse_headings` API and the
  existing `SourcePosition` offsets in `turbovault-core`.

## Open Questions

1. **Parameter naming.** `head_lines`/`tail_lines`, or `offset`/`limit` as used
   by some other tooling? Offset/limit generalises to mid-file ranges;
   head/tail reads more clearly but cannot express "lines 200–250". Happy to
   follow maintainer preference, since this sets the convention.
2. **Combining selectors.** If both positional and structural parameters are
   given — select sections then trim within them, or reject as ambiguous? A
   v1 that rejects the combination is simpler to reason about.
3. **`heading_pattern` semantics.** Full regex, or a simpler prefix/substring
   match? Regex is more expressive but is user input reaching a regex engine.
4. **Streaming.** Should a large-file path avoid reading the whole file? It
   only helps the positional selector — the structural one needs the full
   content to parse headings anyway. Suggest deferring.
5. **Hash on partial reads.** Proposal is to omit it. If maintainers prefer to
   return it for convenience, the `truncated` flag plus a documented warning
   would be the fallback — but omission is the safer default.

## Implementation

I'd like to implement this. I have a branch started (`tail-read`) and the
change is scoped to the two files listed under Impact.

Before writing code I'd welcome guidance on the Open Questions above,
particularly (1) parameter naming and (2) whether to accept or reject
combined selectors — both fix the public API shape, and I'd rather match
your preference than have it changed in review.
## Priority

Medium — a context-efficiency improvement rather than a correctness fix, but the
cost of the current behaviour grows with vault age, and append-only logs are a
common Obsidian pattern.

Helfrid:
Closes #52.

`read_note` gains five optional parameters for reading part of a note instead of
all of it. Omitting all of them leaves the existing whole-file behaviour
byte-for-byte unchanged, `hash` included.

```
head_lines: N       first N lines
tail_lines: N       last N lines
heading_level: 1-6  which heading level delimits a section
heading_equals: S   exact heading text, case-sensitive, no # markers
last_sections: N    last N matching sections
```

Line selectors and heading selectors are mutually exclusive; passing both is a
validation error rather than a guess. Omitting `last_sections` returns every
matching section, which is how a single named section is fetched from anywhere
in a file.

The motivating case is an append-only log with one `##` entry per day: reading
the last three entries of a 463-line note returns 21 lines.

## Answers to the open questions in #52

I asked for guidance before implementing and did not want to block, so these are
the choices I made. Each is cheap to change if you would rather have it
differently.

**1. Parameter naming — counts, not indices.** `head_lines`/`tail_lines` are
counts. A count needs no prior knowledge of the file, whereas expressing "the
last 30 lines" as an index range requires knowing the total line count in
advance, and no existing tool reports it. A range can express "first N" but not
"last N"; that asymmetry decided it. Consequence: mid-file line ranges are not
expressible, which I think is acceptable because `heading_equals` reaches a
named section anywhere in a file and keeps working as the file grows, where a
line number does not.

**2. Combined selectors — rejected.** "The first 10 lines *and* the last 10
lines" has no single defensible answer: two disjoint chunks would stop the
result being a substring and `returned_lines` could not describe it. Silently
picking one interpretation seemed worse than an error.

**3. `heading_pattern` → `heading_equals`.** The issue proposed a regex; this
implements exact match instead. In practice `heading_level` already does the
work a shape-matching pattern would, and `Regex::is_match` is unanchored, so
`heading_pattern: "2026"` would silently also match a heading reading "Notes
about 2026" — a surprise worth avoiding in an LLM-facing API. Rust's `regex`
crate would have been *safe* to expose (finite automata, linear time, no
ReDoS); this was a usability call, not a security one. `heading_contains` could
be added later if a shape match is wanted.

**4. Streaming — deferred.** Section selection needs the whole content to parse
headings, so streaming would only ever help the line selectors. The signature
does not preclude it.

## The hash is deliberately omitted on a partial read

`read_note` returns a `hash` that callers pass back as `write_note`'s
`expected_hash`, becoming `Precondition::ExpectBlob`. That token asserts the
caller has seen the file's current state — false when only part of it was
returned. A caller could read 50 lines, submit those 50 lines as a whole-file
write with a valid hash, and the precondition would pass, silently destroying
the remainder.

Omitting it makes an unhashed `write_note` fall back to `ExpectAbsent`, which
refuses to clobber an existing file. `hash_for_active_backend` is now called
only on the no-selector path, so this is structural rather than a convention to
remember. `edit_note` uses `for_in_place` and needs no token, so it leads
`related` and the `usage` text points at it.

There is an integration test for exactly this: partial read, then an unhashed
whole-file `write_note`, asserted refused with the note intact.

## Where the code lives

- `crates/turbovault-tools/src/file_tools.rs` — `SliceSpec`, `SliceResult`,
  `SectionSlice`, and `slice_content`. Slicing is done on line boundaries by
  byte offset so the result is a verbatim substring: CRLF endings and trailing
  newlines survive, and a line boundary cannot split a UTF-8 codepoint.
- `crates/turbovault/src/tools/providers/files.rs` — the tool parameters,
  response shape and `#[tool]` metadata.

Section slicing uses `turbovault_parser::parse_headings`, whose
`SourcePosition` offsets allow slicing the raw content directly, so a section
is returned verbatim rather than rebuilt from a parsed representation. Using
the parser rather than scanning for lines beginning `#` also means a comment
inside a fenced code block is correctly not treated as a heading.

**Section boundary rule:** a section runs from its heading to the next heading
whose level is numerically less than or equal to its own — a sibling or an
ancestor — or to end of file. So a level-2 section ends at the next level-2 or
level-1 heading and *includes* any level-3 subheadings and their content.
Ending it at the next heading of any kind would truncate a section at its own
first subheading. This is the bulk of the added tests.

## Notes for review

**`tool_catalog.json` is regenerated.** Adding parameters changes `read_note`'s
`inputSchema`, and that fixture is pinned byte-for-byte with no blessing
mechanism, so the 51-line diff there is mechanical. I regenerated it with a
throwaway test which is not part of this PR. If a `--bless` path would be
useful I am happy to add one separately.

**`heading_level` is validated at runtime, not in the schema.** The wire type is
`u8`, so the derived schema advertises `0-255`. Markdown has six levels, so
`SliceSpec::selector` rejects anything outside `1..=6`; without that an
out-of-range level would silently match nothing. Constraining the schema itself
would need a newtype, which seemed heavier than the problem.

**`first_sections` is not exposed as a tool parameter.** With it, `read_note`
reached 8 arguments and clippy's `too_many_arguments` fires at 7; no MCP tool in
this repo currently carries an `allow` for it. It remains on `SliceSpec` for
library callers. Little is lost, since `heading_equals` reaches a named section
anywhere and omitting `last_sections` returns every match, so the only
unavailable phrasing is "the first N sections". Say the word if you would rather
have the `allow` and the symmetry.

**`FileTools::read_file_slice` appears and disappears** across the four commits.
I added it as an async wrapper, then removed it when the tool layer turned out
to need `slice_content` directly: `Ok(None)` carries no content, so a wrapper
would force a second read on the whole-file path. Reviewing the final diff
rather than commit-by-commit avoids the detour.

## Testing

- `cargo test --workspace --all-features --all-targets` green
- `cargo clippy --workspace --all-features --all-targets -- -D warnings` clean
- `cargo fmt --all --check` clean

Unit tests cover both selectors, the boundary rule (sibling, ancestor, nested
subheadings included, last section to EOF), exact-match near-misses including a
trailing space, CRLF and multi-byte content, empty files, counts larger than the
file, and every rejected combination. Integration tests in
`test_mcp_provider_workflows.rs` drive the public MCP facade and assert the
whole-file path is unchanged, that a partial read carries no hash, and that a
following unhashed overwrite is refused.

Also verified by hand against a real 2,400-note vault through both Claude Code
and Claude Desktop.

## Attribution

Planned and implemented with Claude Opus 5. I made the API decisions recorded
above, reviewed every diff, and ran the test suite and clippy locally before
each commit. This is my first contribution here, so I would rather over-explain
the reasoning than have you infer it from the diff.

### DIFF
--- a/crates/turbovault-tools/src/file_tools.rs
+++ b/crates/turbovault-tools/src/file_tools.rs
@@ -106,6 +106,17 @@ impl SliceSpec {
     /// rejected here rather than resolved by guessing: there is no defensible
     /// answer to "the first 10 lines *and* the last 10 lines".
     fn selector(&self) -> Result<Option<Selector>> {
+        // Markdown has six heading levels. The wire type is `u8`, so the
+        // generated JSON schema advertises 0-255; reject the rest here rather
+        // than silently matching nothing.
+        if let Some(level) = self.heading_level
+            && !(1..=6).contains(&level)
+        {
+            return Err(Error::validation_error(format!(
+                "heading_level must be between 1 and 6, got {level}"
+            )));
+        }
+
         let take = match (self.first_sections, self.last_sections) {
             (Some(_), Some(_)) => {
                 return Err(Error::validation_error(
@@ -356,21 +367,6 @@ impl FileTools {
     pub async fn read_file(&self, path: &str) -> Result<String> {
         let file_path = PathBuf::from(path);
         self.manager.read_file(&file_path).await
-    }
-
-    /// Read part of a file from the vault.
-    ///
-    /// Reads the full raw content through [`Self::read_file`], inheriting its
-    /// frontmatter-preserving guarantee, then applies `spec`. Returns
-    /// `Ok(None)` when `spec` selects nothing, so the caller can fall through
-    /// to the unchanged whole-file path.
-    pub async fn read_file_slice(
-        &self,
-        path: &str,
-        spec: &SliceSpec,
-    ) -> Result<Option<SliceResult>> {
-        let content = self.read_file(path).await?;
-        slice_content(&content, spec)
     }
 
     /// Write a file to the vault with mode support (creates directories as needed)
@@ -1528,6 +1524,33 @@ Did: c
     }
 
     #[test]
+    fn test_slice_rejects_out_of_range_heading_level() {
+        for level in [0u8, 7, 200] {
+            let spec = SliceSpec {
+                heading_level: Some(level),
+                ..Default::default()
+            };
+            let err = slice_content(LOG, &spec).unwrap_err();
+            assert!(
+                err.to_string()
+                    .contains("heading_level must be between 1 and 6"),
+                "level {level}: {err}"
+            );
+        }
+    }
+
+    #[test]
+    fn test_slice_accepts_all_valid_heading_levels() {
+        for level in 1u8..=6 {
+            let spec = SliceSpec {
+                heading_level: Some(level),
+                ..Default::default()
+            };
+            assert!(slice_content(LOG, &spec).is_ok(), "level {level}");
+        }
+    }
+
+    #[test]
     fn test_slice_rejects_first_and_last_sections_together() {
         let spec = SliceSpec {
             first_sections: Some(1),
@@ -1551,104 +1574,5 @@ Did: c
             );
             assert_eq!(r.sections_matched, Some(0), "probe: {probe:?}");
         }
-    }
-
-    // ---- partial reads through the vault ----
-
-    fn slice_manager(vault_dir: &std::path::Path) -> Arc<VaultManager> {
-        use turbovault_core::{ServerConfig, VaultConfig};
-        let mut config = ServerConfig::new();
-        config
-            .vaults
-            .push(VaultConfig::builder("test", vault_dir).build().unwrap());
-        Arc::new(VaultManager::new(config).unwrap())
-    }
-
-    /// `read_file_slice` must return the same bytes `read_file` would, for the
-    /// portion selected — including the frontmatter `read_file` preserves.
-    #[tokio::test]
-    async fn test_read_file_slice_tail() {
-        let temp = tempfile::TempDir::new().unwrap();
-        std::fs::write(
-            temp.path().join("log.md"),
-            "---\ntitle: Log\n---\n\n## First\nbody\n\n## Second\nmore\n",
-        )
-        .unwrap();
-        let tools = FileTools::new(slice_manager(temp.path()));
-
-        let result = tools
-            .read_file_slice(
-                "log.md",
-                &SliceSpec {
-                    tail_lines: Some(2),
-                    ..Default::default()
-                },
-            )
-            .await
-            .unwrap()
-            .unwrap();
-
-        assert_eq!(result.content, "## Second\nmore\n");
-        assert_eq!(result.total_lines, 9);
-        assert!(result.truncated);
-    }
-
-    #[tokio::test]
-    async fn test_read_file_slice_sections() {
-        let temp = tempfile::TempDir::new().unwrap();
-        std::fs::write(temp.path().join("log.md"), LOG).unwrap();
-        let tools = FileTools::new(slice_manager(temp.path()));
-
-        let result = tools
-            .read_file_slice(
-                "log.md",
-                &SliceSpec {
-                    heading_level: Some(2),
-                    last_sections: Some(1),
-                    ..Default::default()
-                },
-            )
-            .await
-            .unwrap()
-            .unwrap();
-
-        assert_eq!(result.content, "## 2026-08-22\nDid: c\n");
-        assert_eq!(result.sections_matched, Some(3));
-    }
-
-    /// No selector: `None`, so the caller keeps using the whole-file path.
-    #[tokio::test]
-    async fn test_read_file_slice_without_selector_returns_none() {
-        let temp = tempfile::TempDir::new().unwrap();
-        std::fs::write(temp.path().join("log.md"), LOG).unwrap();
-        let tools = FileTools::new(slice_manager(temp.path()));
-
-        let result = tools
-            .read_file_slice("log.md", &SliceSpec::default())
-            .await
-            .unwrap();
-        assert!(result.is_none());
-    }
-
-    /// A missing file must surface as an I/O error, not an empty slice.
-    #[tokio::test]
-    async fn test_read_file_slice_propagates_read_error() {
-        let temp = tempfile::TempDir::new().unwrap();
-        let tools = FileTools::new(slice_manager(temp.path()));
-
-        let err = tools
-            .read_file_slice(
-                "nope.md",
-                &SliceSpec {
-                    tail_lines: Some(1),
-                    ..Default::default()
-                },
-            )
-            .await
-            .unwrap_err();
-        assert!(
-            matches!(err, Error::Io(_) | Error::FileNotFound { .. }),
-            "got: {err:?}"
-        );
     }
 }
--- a/crates/turbovault-tools/src/file_tools.rs
+++ b/crates/turbovault-tools/src/file_tools.rs
@@ -358,6 +358,21 @@ impl FileTools {
         self.manager.read_file(&file_path).await
     }
 
+    /// Read part of a file from the vault.
+    ///
+    /// Reads the full raw content through [`Self::read_file`], inheriting its
+    /// frontmatter-preserving guarantee, then applies `spec`. Returns
+    /// `Ok(None)` when `spec` selects nothing, so the caller can fall through
+    /// to the unchanged whole-file path.
+    pub async fn read_file_slice(
+        &self,
+        path: &str,
+        spec: &SliceSpec,
+    ) -> Result<Option<SliceResult>> {
+        let content = self.read_file(path).await?;
+        slice_content(&content, spec)
+    }
+
     /// Write a file to the vault with mode support (creates directories as needed)
     ///
     /// `message` is the commit subject the underlying `VaultManager` write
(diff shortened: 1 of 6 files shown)


nicholasjpaterno:
Merging, and thanks for the thorough write-up on the open questions. Answering them in the PR rather than waiting on me was the right instinct.

The detail I want to call out is refusing to return a `hash` on a partial read. That token asserts the caller has seen the file's current state, and handing it to `write_note` after reading forty lines of a four-hundred-line note would let a whole-file overwrite satisfy its precondition and destroy the rest. Plenty of implementations would have returned it without thinking. Spelling the reason out in the tool description too means an agent can actually act on it.

Counts over indices is the right call for the same reason you gave: `heading_equals` keeps working as a file grows and a line number does not.

Worth knowing that this PR never got CI, fork branches report no checks on this repo. I verified locally against current main: 1211 tests pass with your change merged alongside the rest of the backlog, clippy and fmt clean.

